PR #178, opened 2023-11-02
fix for delivered_amount

RichardAH:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell8:
The changes in `DeliveredAmount.cpp` involve an update to the condition that determines when to return the `sfAmount` field. Previously, the `sfAmount` field was returned if the ledger index was greater than or equal to 4594095, or if the close time was greater than a specific timestamp (February 2014). The update adds a new condition that also checks if the `serializedTx` is present and contains the `sfNetworkID` field. If this new condition is met, the `sfAmount` field will be returned as well.

PR #210, opened 2023-11-18
Update Import.cpp

wojake:
Include Import tx fee into Leader Header change for XAH supply

The current layout does not include the Import tx fee, which should as it is technically minted and then consumed as the tx fee. While it is not introduced economically, it is technically introduced to the supply.

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/app/tx/impl/Import.cpp
+++ b/src/ripple/app/tx/impl/Import.cpp
@@ -1303,7 +1303,8 @@ Import::doApply()
     }
 
     // update the ledger header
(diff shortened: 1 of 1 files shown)


wojake:
I think this needs to be an amendment

sublimator:
that was fast

wojake:
I was wrong :P

PR #220, opened 2023-11-24
FXV1: Namespace Limit

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1488,6 +1488,20 @@ DEFINE_HOOK_FUNCTION(
...
+
+    if (!sleAccount)
+        return tefINTERNAL;
+
+    if (view.rules().enabled(fixXahauV1))
+    {
+        STVector256 vec = sleAccount->getFieldV256(sfHookNamespaces);
+        if (vec.size() > 256)
+        {
+            return OUT_OF_BOUNDS;
```
RichardAH: I think we should add a new error code for this. TOO_MANY_NAMESPACES

RichardAH (changes requested):

src/ripple/app/hook/impl/applyHook.cpp
dangell7: Done.

PR #221, opened 2023-11-28
Fix governance vote purge

### LINKED ISSUE #211: Governance hook: Votes of removed members are not purged
## Issue Description
When Table members are removed from the table, their votes are still counted in the vote count. 

## Steps to Reproduce
- Cast votes from L2 table members
- Replace members with new ones 
For example table with 2 seats: 0 and 1
0 seat vote to place rNewAddress0 to place 0 (its own place)
0 seat vote to place rNewAddress1 to place 1 
1 seat vote to place rNewAddress0 to place 0 (0 seat gets replaced)
1 seat vote to place rNewAddress1 to place 1 (its own place) (1st seat gets replaced)

## Expected Result
I would expect the votes from removed members not to be included in the Votes Count. 

## Actual Result
The vote list and the vote count keep the votes from removed members

## Environment
Hook: 78CA3F5BD3D4F7B32A6BEBB3844380A9345C9BA496EFEB30314BDDF405D7B4B3
Testnet

## Supporting Files
https://test.xahauexplorer.com/governance/roUo3ygV92bdhfE1v9LGpPETXvJv2kQv5

tequdev:

Fixed `previous_member` and `vote_key` are set correctly when deleting votes in Governance Hook.

### Context of Change

- fix #211

The implementation didn't match the top comment in the govern.c file.
- L1/L2 information was missing when obtaining State.

Fixed existence comparison conditions.

### Type of Change

- [x ] Bug fix (non-breaking change which fixes an issue)

If each governance table continues to use this governance Hook, it is recommended that the Hook be updated.

## Test Plan

Since the author is not familiar with C++ code and cannot write sufficient test code, so used JS for the test instead.

https://github.com/tequdev/xahau-governance-hook-test


hook/genesis/govern.c
```diff
@@ -637,43 +637,48 @@ int64_t hook(uint32_t r)
...
-                    previous_member[1] = i < 2 ? 'R' : i < 12 ? 'H' : 'S';
-                    previous_member[2] = 
-                        i == 0 ? 'R' :
-                        i == 1 ? 'D' :
-                        i < 12 ? i - 2 :
-                        i - 12;
-
-                    uint8_t vote_key[32];
-                    if (state(SBUF(vote_key), SBUF(previous_member)) == 32)
+                    for (int i = 0; GUARD(66), i < 32; ++i)
```
tequdev, line 642: I could understand the other code changes, but why is 66?

dangell7: I have to revisit that during testing. I had 64 but it complained.

tequdev: Ah, Okay.

Looks like this part of the documentation needs to be correct.
https://xrpl-hooks.readme.io/docs/loops-and-guarding#nested-loops

dangell7: Yeah gonna wait till I confirm it wasnt a mistake on my part then I will update. I'll keep this conversation open for now.

hook/genesis/govern.c
```diff
@@ -637,43 +637,48 @@ int64_t hook(uint32_t r)
...
+                        previous_member[1] = i < 2 ? 'R' : i < 12 ? 'H' : 'S';
+                        previous_member[2] = 
+                            i == 0 ? 'R' :
+                            i == 1 ? 'D' :
+                            i < 12 ? i - 2 :
+                            i - 12;
+                        previous_member[3] = tbl;
+
+                        uint8_t vote_key[32];
+                        if (state(SBUF(vote_key), SBUF(previous_member)) > 0)
```
tequdev: This section seems to have to be my following code.
`if (state(vote_key + 12, 20, SBUF(previous_member)) == 20)`

if `if (state(SBUF(vote_key), SBUF(previous_member)) > 0)`
```
previous_member==:: 565310010000000000000000AE123A8556F3CF91154711376AFB0F894F832B3D
vote_key==:: 4353100156F3CF91154711376AFB0F894F832B3D000000000000000000000000
```

if `if (state(vote_key + 12, 20, SBUF(previous_member)) == 20)`,
```
previous_member==:: 565310010000000000000000AE123A8556F3CF91154711376AFB0F894F832B3D
vote_key==:: 435310010000000000000000AE123A8556F3CF91154711376AFB0F894F832B3D
```

---

It looks like I can leave `if (state(SBUF(vote_key), SBUF(previous_member)) > 0)` as is and just rewrite the first few bytes when vote_key matches previous_member's value, but I don't know why.

tequdev: `state(SBUF(vote_key), SBUF(previous_member))` returns 20 not 32

it should be 32

tequdev: but `sizeof(previous_member)` is 32
so `state(SBUF(vote_key), SBUF(previous_member))` = `state(vote_key, 32, previous_member, 32)`

dangell7: It's the size of the data being written to the buffer. We will need to use special padding it looks like.

It can be 32 bytes for a hook, 8 bytes for an xfl or 20 bytes for an seat.

hook/genesis/govern.c
```diff
@@ -650,7 +650,14 @@ int64_t hook(uint32_t r)
                         previous_member[3] = tbl;
 
                         uint8_t vote_key[32];
```
tequdev: `uint8_t vote_key[32] = {};`

and

---

In the code `uint8_t vote_key[32];`, the value of the previous `vote_key` was used in the loop process without being initialized.

hook/genesis/govern.c
```diff
@@ -650,7 +650,14 @@ int64_t hook(uint32_t r)
...
-                        if (state(SBUF(vote_key), SBUF(previous_member)) > 0)
+                        uint8_t ts = 
+                            previous_member[1] == 'H' ? 32 :      // hook topics are a 32 byte hook hash
+                            previous_member[1] == 'S' ? 20 :      // account topics are a 20 byte account ID
+                            8;                        // reward topics are an 8 byte le xfl
+                        
+                        uint8_t padding = 32 - ts;
+
+                        if (state(vote_key + padding, 32, SBUF(previous_member)) == 32)
```
tequdev: `if (state(vote_key + padding, ts, SBUF(previous_member)) == ts)`

works for me

RichardAH:
I think this is correct but we really need a deterministic build setup for these "embedded" hooks so we can prove what the wasm binary contains. Not that I don't trust @tequdev, but because no one should have to trust an opaque binary to begin with. I feel uncomfortable about merging opaque binaries, even ones I generate myself.

dangell7:
> I think this is correct but we really need a deterministic build setup for these "embedded" hooks so we can prove what the wasm binary contains. Not that I don't trust @tequdev, but because no one should have to trust an opaque binary to begin with. I feel uncomfortable about merging opaque binaries, even ones I generate myself.

So you want to build and store the hex binary in the repo during the git actions? We could also do it how you did the tests where it shows both the c and the hex?

https://raw.githubusercontent.com/Xahau/xahaud/refs/heads/dev/src/test/app/SetHook_wasm.h

RichardAH:
> > I think this is correct but we really need a deterministic build setup for these "embedded" hooks so we can prove what the wasm binary contains. Not that I don't trust @tequdev, but because no one should have to trust an opaque binary to begin with. I feel uncomfortable about merging opaque binaries, even ones I generate myself.
> 
> So you want to build and store the hex binary in the repo during the git actions? We could also do it how you did the tests where it shows both the c and the hex?
> 
> https://raw.githubusercontent.com/Xahau/xahaud/refs/heads/dev/src/test/app/SetHook_wasm.h

Yeah I think automated building of the blobs during the git actions is the correct answer. But they have to be deterministic builds.

PR #223, opened 2023-11-29
Add HookParameters fee calc to all txs

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -350,6 +350,23 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
...
+        for (auto const& param : params)
+        {
+            paramBytes += (param.isFieldPresent(sfHookParameterName)
+                               ? param.getFieldVL(sfHookParameterName).size()
+                               : 0) +
+                (param.isFieldPresent(sfHookParameterValue)
+                     ? param.getFieldVL(sfHookParameterValue).size()
+                     : 0);
+        }
+        accumulator += XRPAmount{static_cast<XRPAmount>(paramBytes)};
```
RichardAH, line 367: these used to have overflow checks didn't they? what happened to that logic?

RichardAH (changes requested):

src/ripple/app/tx/impl/Transactor.cpp
dangell7: The overflow checks were in SetHook.cpp but not in Invoke.cpp. I can add them here though

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -566,40 +566,6 @@ SetHook::calculateBaseFee(ReadView const& view, STTx const& tx)
             createFee = XRPAmount{hook::computeCreationFee(
                 hookSetObj.getFieldVL(sfCreateCode).size())};
 
-        XRPAmount paramFee{0};
```
RichardAH: this code is adding up the cost to submit the txn based on the hook parameters inside the hooks array inside hookset, this is a different bill and different parameters to the top level hook parameters array so it should be left in

dangell7: Update, thank you!

RichardAH (approved):

PR #224, opened 2023-11-30
FXV1: Optional Offer Sequence

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/app/tx/impl/CancelOffer.cpp
+++ b/src/ripple/app/tx/impl/CancelOffer.cpp
@@ -39,10 +39,26 @@ CancelOffer::preflight(PreflightContext const& ctx)
         return temINVALID_FLAG;
     }
 
-    if (!ctx.tx[sfOfferSequence])
+    if (ctx.rules.enabled(fixXahauV1))
     {
-        JLOG(ctx.j.trace()) << "CancelOffer::preflight: missing sequence";
-        return temBAD_SEQUENCE;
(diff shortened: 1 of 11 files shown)


src/ripple/app/tx/impl/CancelOffer.cpp
```diff
@@ -39,10 +39,26 @@ CancelOffer::preflight(PreflightContext const& ctx)
...
+        {
+            JLOG(ctx.j.trace())
+                << "CancelOffer::preflight: invalid sequence or offer id";
+            return temBAD_SEQUENCE;
+        }
+    }
+    else
+    {
+        if (!ctx.tx.isFieldPresent(sfOfferSequence) ||
+            ctx.tx[sfOfferSequence] == 0)
```
RichardAH, line 57: condition should match exactly what it was `if (!ctx.tx[sfOfferSequence])` otherwise if the behaviour is even slightly different this can fork

RichardAH: ensure the behaviour without the amendment is identical in the apply too

RichardAH (changes requested):

src/ripple/app/tx/impl/CancelOffer.cpp
dangell7: This was written like this because `sfOfferSequence` is now optional. So I think `ctx.tx[sfOfferSequence]` fails hard. I can revert and show the output.

PR #226, opened 2023-12-01
Build on most cores (nproc / 3 » nproc / 1.337)

WietseWind:
Use those resources! 🎉🚀

Build was at nproc / 3, probably to keep some system resources available for other builds. But as we're not building in parallel on the same machine, let's just use those resources (minus a little bit for other system processes).


alloynetworks (approved):
LGTM. Given our specs.

dangell7 (approved):

PR #261, opened 2024-01-19
Fix: TSH Updates & Emitted Txn

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


tequdev (commented):
Some sections (ttTRUST_SET,ttNFTOKEN_CREATE/ACCEPT_OFFER,ttPAYCHAN_CLAIM, etc.) seem to be replacing canRollback with tshWEAK instead of tshSTRONG without fixXahauV2 checking. Is this OK?

RichardAH:
> Some sections (ttTRUST_SET,ttNFTOKEN_CREATE/ACCEPT_OFFER,ttPAYCHAN_CLAIM, etc.) seem to be replacing canRollback with tshWEAK instead of tshSTRONG without fixXahauV2 checking. Is this OK?

canRollback was derived from a static table in applyHook.h, so all these replacements should be identical to what they already were in that table, but please feel free to check

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -228,16 +254,16 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
 
             if (bo)
             {
-                ADD_TSH(bo->getAccountID(sfOwner), canRollback);
+                ADD_TSH(bo->getAccountID(sfOwner), tfStrongTSH);
```
RichardAH: fix this

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -228,16 +254,16 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
...
                 if (bo->isFieldPresent(sfDestination))
-                    ADD_TSH(bo->getAccountID(sfDestination), canRollback);
+                    ADD_TSH(bo->getAccountID(sfDestination), tshWEAK);
             }
 
             if (so)
             {
-                ADD_TSH(so->getAccountID(sfOwner), canRollback);
+                ADD_TSH(so->getAccountID(sfOwner), tfStrongTSH);
```
RichardAH: here too

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -253,24 +279,21 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
                 auto const offer = getNFTOffer(offerID, rv);
                 if (offer)
                 {
-                    ADD_TSH(offer->getAccountID(sfOwner), canRollback);
+                    ADD_TSH(offer->getAccountID(sfOwner), tfStrongTSH);
```
RichardAH: and here

PR #278, opened 2024-02-02
Amendment: `featureRemit`

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


tequdev:
Is it expected that the transaction will succeed if none of the fields sfAmounts, sfURITokenIDs, or sfMintURIToken are specified?

From the code it looks like it will succeed, but it may be appropriate that it will be an error.

It's okay if success is expected.

dangell7:
> Is it expected that the transaction will succeed if none of the fields sfAmounts, sfURITokenIDs, or sfMintURIToken are specified?
> 
> From the code it looks like it will succeed, but it may be appropriate that it will be an error.
> 
> It's okay if success is expected.

It will create the account if the account does not exist.

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -70,6 +70,17 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
...
     {
+
+        case ttREMIT: {
+            if (destAcc)
+                ADD_TSH(*destAcc, tshSTRONG);
+
+            if (tx.isFieldPresent(sfInform))
+                ADD_TSH(tx.getAccountID(sfInform), tshWEAK);
+
+            break;
```
tequdev: If URIToken is included, shouldn't the issuer be set to TSH?

dangell7: Issuer is the OTXN Account and so by default, the strong TSH.

dangell7: Yes it should. Thank you

tequdev (commented):
If the account specified for sfInform is a Destination or an issuing account with a Burnable URIToken issued by a third party, will the Inform account fire the Hook as Strong or Weak?

I don't know how cases where the same account is registered multiple times with ADD_TSH are processed.

dangell7:
> If the account specified for sfInform is a Destination or an issuing account with a Burnable URIToken issued by a third party, will the Inform account fire the Hook as Strong or Weak?
> 
> I don't know how cases where the same account is registered multiple times with ADD_TSH are processed.

I'm more in favor of removing inform. It's not allowed to be the account or destination. I will fix this. 

It would fire as strong, if set to the issuer, with burnable, else weak I believe, I will double check this.

src/ripple/protocol/impl/InnerObjectFormats.cpp
```diff
@@ -143,6 +143,21 @@ InnerObjectFormats::InnerObjectFormats()
             {sfPublicKey, soeREQUIRED},
             {sfAccount, soeOPTIONAL},
         });
+
+    add(sfAmountEntry.jsonName.c_str(),
+        sfAmountEntry.getCode(),
+        {
+            {sfAmount, soeREQUIRED},
+            {sfFlags, soeOPTIONAL},
```
tequdev: What is it used for?

dangell7: Informing a 3rd party

dangell7: I read that wrong, inform is for 3rd party. The flag here is unused. I think we were discussing the option and never removed it. I will remove it now.

src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -116,6 +116,21 @@ TxFormats::TxFormats()
...
+        {
+            {sfDestination, soeREQUIRED},
+            {sfAmounts, soeOPTIONAL},
+            {sfURITokenIDs, soeOPTIONAL},
+            {sfMintURIToken, soeOPTIONAL},
+            {sfInvoiceID, soeOPTIONAL},
+            {sfDestinationTag, soeOPTIONAL},
+            {sfTicketSequence, soeOPTIONAL},
+            {sfBlob, soeOPTIONAL},
+            {sfInform, soeOPTIONAL},
```
tequdev, line 130: Do you Keep sfInform Field?

src/ripple/protocol/impl/InnerObjectFormats.cpp
tequdev: Do you keep sfFlags in sfAmountEntry?

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+    XRPAmount nativeRemit{0};
+
+    AccountID const dstAccID{ctx_.tx[sfDestination]};
+    auto sleDstAcc = sb.peek(keylet::account(dstAccID));
+    auto const flags = !sleDstAcc ? 0 : sleDstAcc->getFlags();
+
+    // Check if the destination has disallowed incoming
+    if (sb.rules().enabled(featureDisallowIncoming) &&
+        (flags & lsfDisallowIncomingRemit))
+        return tecNO_PERMISSION;
```
tequdev, line 276: Is lsfDisallowIncomingRemit flag finally opt-out? I'd like to know the reason?

src/ripple/protocol/impl/TxFormats.cpp
dangell7: Yes Inform will stay.

src/ripple/protocol/impl/InnerObjectFormats.cpp
dangell7: No I will remove this.

src/ripple/app/tx/impl/Remit.cpp
dangell7: Yes it is opt out. The goal of ttRemit is so that the receiving user doesn't have to perform any action.

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012, 2013 Ripple Labs Inc.
```
RichardAH, line 4: these copyrights are wrong and need to be updated in a number of places, but this PR isn't that place

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+#include <ripple/basics/Log.h>
+#include <ripple/ledger/View.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Indexes.h>
+namespace ripple {
+
+TxConsequences
+Remit::makeTxConsequences(PreflightContext const& ctx)
+{
+    return TxConsequences{ctx.tx, TxConsequences::normal};
```
RichardAH: this is probably wrong, see the payment transactor tx consequences https://github.com/Xahau/xahaud/blob/caa8b382d8dfbf3bfd14e183e2e9cf10e68ffc0a/src/ripple/app/tx/impl/Payment.cpp#L31-L44

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+    AccountID const dstID = ctx.tx.getAccountID(sfDestination);
+    AccountID const srcID = ctx.tx.getAccountID(sfAccount);
+
+    if (dstID == srcID)
+    {
+        JLOG(ctx.j.warn()) << "Malformed transaction: Remit to self.";
+        return temREDUNDANT;
+    }
+
+    if (ctx.tx.getFieldVL(sfBlob).size() > (128 * 1024))
```
RichardAH: sfBlob is optional but that's not checked here, should be if (ctx.tx.isFieldPresent(sfBlob) &&...

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+                           << ctx.tx.getTransactionID();
+        return temMALFORMED;
+    }
+
+    // sanity check amounts
+    if (ctx.tx.isFieldPresent(sfAmounts))
+    {
+        if (ctx.tx.getFieldArray(sfAmounts).size() > 32)
+        {
+            JLOG(ctx.j.warn()) << "Malformed: AmountEntrys Exceed Limit `32`.";
```
RichardAH: AmountEntry count exceeds

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+        }
+    }
+
+    // sanity check minturitoken
+    if (ctx.tx.isFieldPresent(sfMintURIToken))
+    {
+        STObject const& mint = const_cast<ripple::STTx&>(ctx.tx)
+                                   .getField(sfMintURIToken)
+                                   .downcast<STObject>();
+        // RH TODO: iterate mint fields detect any that shouldnt be there
```
RichardAH: should be done like here: https://github.com/Xahau/xahaud/blob/caa8b382d8dfbf3bfd14e183e2e9cf10e68ffc0a/src/ripple/app/tx/impl/SetHook.cpp#L716-L732

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+        Blob const uri = mint.getFieldVL(sfURI);
```
RichardAH, line 183: not sure how well inner object template system works, we should probably guard this with a isFieldPresent to be safe

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+        {
+            if (mint.getFieldU32(sfFlags) & tfURITokenMintMask)
+                return temINVALID_FLAG;
+        }
+    }
+
+    // sanity check uritokenids
+    if (ctx.tx.isFieldPresent(sfURITokenIDs))
+    {
+        if (ctx.tx.getFieldV256(sfURITokenIDs).size() > 32)
```
RichardAH: Not sure if vec256 size 0 is possible but we should probably defensively check here

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+    auto sleDstAcc = sb.peek(keylet::account(dstAccID));
+    auto const flags = !sleDstAcc ? 0 : sleDstAcc->getFlags();
+
+    // Check if the destination has disallowed incoming
+    if (sb.rules().enabled(featureDisallowIncoming) &&
+        (flags & lsfDisallowIncomingRemit))
+        return tecNO_PERMISSION;
+
+    // Check if the destination account requires deposit authorization.
+    bool const depositAuth{view().rules().enabled(featureDepositAuth)};
```
RichardAH: should be sb.rules(). one view only in use at a time

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+    if (depositAuth && sleDstAcc && (sleDstAcc->getFlags() & lsfDepositAuth))
+    {
+        if (!view().exists(keylet::depositPreauth(dstAccID, srcAccID)))
```
RichardAH: sb.exists

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+        sleMint->setFieldVL(sfURI, mintURI);
+
+        if (mint.isFieldPresent(sfDigest))
+            sleMint->setFieldH256(sfDigest, mint.getFieldH256(sfDigest));
+
+        sleMint->setFieldU32(
+            sfFlags,
+            mint.isFieldPresent(sfFlags) ? mint.getFieldU32(sfFlags) : 0);
+
+        auto const page = view().dirInsert(
```
RichardAH: should be sb.dirInsert?

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,545 @@
...
+                     canXfer != tesSUCCESS)
+                return canXfer;
+
+            // compute the amount the source will need to send
+            // in remit the sender pays all transfer fees, so that
+            // the destination can always be assured they got the exact amount
+            // specified. therefore we need to compute the amount + transfer fee
+            auto const srcAmt =
+                issuerAccID != srcAccID && issuerAccID != dstAccID
+                ? multiply(amount, transferRate(sb, issuerAccID))
```
RichardAH, line 523: I don't think this transfer rate is tested anywhere? given history of getting this wrong please add test cases for 0 xfer rate and non-zero as well

RichardAH (changes requested):

src/ripple/app/tx/impl/Remit.cpp
dangell7: https://github.com/Xahau/xahaud/issues/282

src/ripple/app/tx/impl/Remit.cpp
dangell7: I tested it the same way we did in URIToken after the fix. https://github.com/Xahau/xahaud/blob/4e8c0b4388eaac006e71dd7dd9bca4a0c73e5d5c/src/test/app/Remit_test.cpp#L2017

Will add test for 0 and non 0.

src/ripple/app/tx/impl/Remit.cpp
dangell7: The submit throws if you don't include the URI field. Also throws if you include a field not present in the template. Either way I still added the defensive checks.

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -70,6 +70,39 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
 
     switch (tt)
     {
+        case ttREMIT: {
+            if (destAcc)
+                ADD_TSH(*destAcc, tshSTRONG);
+
+            if (tx.isFieldPresent(sfInform))
```
RichardAH, line 77: We should check if the sfInform is the same as destination or origin

We should also check if the sfInform exists (in doApply) and it should be an error if it does not.

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -70,6 +70,39 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
...
+                        continue;
+
+                    auto const ut = rv.read(id);
+                    if (!ut ||
+                        ut->getFieldU16(sfLedgerEntryType) != ltURI_TOKEN)
+                        continue;
+
+                    auto const owner = ut->getAccountID(sfOwner);
+                    auto const issuer = ut->getAccountID(sfIssuer);
+                    if (issuer != owner)
```
RichardAH: the issuer might also be the same as the destination

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,560 @@
...
+    auto const flags = !sleDstAcc ? 0 : sleDstAcc->getFlags();
+
+    // Check if the destination has disallowed incoming
+    if (sb.rules().enabled(featureDisallowIncoming) &&
+        (flags & lsfDisallowIncomingRemit))
+        return tecNO_PERMISSION;
+
+    // Check if the destination account requires deposit authorization.
+    bool const depositAuth{sb.rules().enabled(featureDepositAuth)};
+    if (depositAuth && sleDstAcc && (sleDstAcc->getFlags() & lsfDepositAuth))
```
RichardAH: sleDstAcc->getFlags() should just be flags, it's set above

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,560 @@
...
+XRPAmount
+Remit::calculateBaseFee(ReadView const& view, STTx const& tx)
+{
+    XRPAmount extraFee{0};
+
+    if (tx.isFieldPresent(sfBlob))
+        extraFee +=
+            XRPAmount{static_cast<XRPAmount>(tx.getFieldVL(sfBlob).size())};
+
+    // RH TODO: add fees
```
RichardAH: not sure what other fees are left to add, this comment can probably go away

RichardAH (changes requested):

src/ripple/app/tx/impl/Remit.cpp
```diff
@@ -0,0 +1,592 @@
...
+            // if the target trustline doesn't exist we need to create it and
+            // pay its reserve
+            if (issuerAccID != dstAccID &&
+                !sb.exists(
+                    keylet::line(dstAccID, issuerAccID, amount.getCurrency())))
+                nativeRemit += objectReserve;
+
+            // action the transfer
+            if (TER result =
+                    accountSend(sb, srcAccID, dstAccID, amount, j, false);
```
RichardAH: should this be (..., true)?

RichardAH (approved):

Issue #288, opened 2024-03-13
Weak TSH accounts don't always show up in metadata


RichardAH:
add a touch timestamp/counter to account root

tequdev:
@RichardAH Can we close this issue?

PR #294, opened 2024-03-15
Touch amendment

RichardAH:
An amendment to ensure that all accounts that are affected or could be affected by a txn appear in the txn metadata, akin to the touch command in *nix systems.

WIP

https://github.com/Xahau/xahaud/issues/288

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell7:
Because the new field `TouchCounter` doesnt exist and is optional the first update leaves us with an empty json `PreviousFields`.

This occurs [here](https://github.com/Xahau/xahaud/blob/acd455f5dfa11b2930a8066d72aef5d67ac52270/src/ripple/ledger/impl/ApplyStateTable.cpp#L172). I dont believe we need do anything I just wanted to notate it.

```
{
    "ModifiedNode" : 
    {
        "FinalFields" : 
        {
            "Account" : "rPMh7Pi9ct699iZUTWaytJUoHcJ7cgyziK",
            "AccountIndex" : "1",
            "Balance" : "997999990",
            "Flags" : 41943040,
            "OwnerCount" : 1,
            "Sequence" : 3,
            "TouchCount" : "1"
        },
        "LedgerEntryType" : "AccountRoot",
        "LedgerIndex" : "DE3BE7FDF6864FB024807B36BFCB4607E7CDA7D4C155C7AFB4B0973D638938BF",
        "PreviousFields" : {}
    }
}
```

dangell7:
For Reviewers:

When a trustline is frozen it updates the `TouchCount` for the account that has the trustline. This didn't occur previously. 

This was one of the only locations where the number of affected nodes is tested. This change does mean that every TSH will get a TouchCount update which may add txns to their account_tx stream which they may or may not understand.

RichardAH:
I think this is basically ready

src/ripple/protocol/Feature.h
```diff
@@ -362,6 +362,8 @@ extern uint256 const fix240819;
 extern uint256 const fixPageCap;
 extern uint256 const fix240911;
 extern uint256 const fixFloatDivide;
+extern uint256 const featureTouch;
+
```
tequdev:
```suggestion
```

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -1079,6 +1079,31 @@ Transactor::checkMultiSign(PreclaimContext const& ctx)
 
 //------------------------------------------------------------------------------
 
+// increment the touch counter on an account
+static void
+touchAccount(
+    ApplyView& view,
+    AccountID const& id,
+    std::optional<std::shared_ptr<SLE>> sle)
```
tequdev: What is the intent of this argument that is not actually used?

tequdev (approved):

src/ripple/app/tx/impl/Transactor.cpp
dangell7: Which argument?

dangell7: @RichardAH I removed this. Can you re-review? Was this done so that later you could call it from another location and include the sle? Can we just remove that and add it later if we need to?

RichardAH: It was forward thinking for other touches where the SLE is already provided by the calling function.
Speaking of which, I just did a quick code review and don't we need touch in a few other places not just unfunded offers?

zgrguric:
Bump.

PR #295, opened 2024-03-15
Add RPC Tests

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/rpc/handlers/LedgerEntry.cpp
+++ b/src/ripple/rpc/handlers/LedgerEntry.cpp
@@ -24,6 +24,7 @@
 #include <ripple/net/RPCErr.h>
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/protocol/Indexes.h>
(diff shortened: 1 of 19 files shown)


src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -231,6 +232,22 @@ doLedgerEntry(RPC::JsonContext& context)
...
+        if (!context.params[jss::public_key].isString())
+        {
+            uNodeIndex = beast::zero;
+            jvResult[jss::error] = "malformedRequest";
+        }
+        else
+        {
+            auto const pkHex =
+                strUnHex(context.params[jss::public_key].asString());
+            auto const pk = PublicKey(makeSlice(*pkHex));
```
RichardAH: I believe the publickey constructor will crash the application if the public key isn't valid, so this is an attack vector.

the pk validity needs to be checked prior to construction

PR #300, opened 2024-03-18
Fix: Missing Headers - Linker Errors

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/protocol/STValidation.h
+++ b/src/ripple/protocol/STValidation.h
@@ -171,7 +171,8 @@ STValidation::STValidation(
 {
     if (checkSignature && !isValid())
     {
-        JLOG(debugLog().error()) << "Invalid signature in validation";
+        JLOG(debugLog().error()) << "Invalid signature in validation: "
+                                 << getJson(JsonOptions::none);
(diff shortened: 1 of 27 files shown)


RichardAH:
Some of the changes here remove JSON dumps from important logging lines. Can you rework it so those dumps remain?

src/ripple/protocol/STValidation.h
```diff
@@ -171,8 +171,7 @@ STValidation::STValidation(
 {
     if (checkSignature && !isValid())
     {
-        JLOG(debugLog().error()) << "Invalid signature in validation: "
-                                 << getJson(JsonOptions::none);
+        JLOG(debugLog().error()) << "Invalid signature in validation";
```
RichardAH: I still want the hex if not the json dumped under this condition, otherwise we wouldnt know which validation/signature was invalid if we had to debug with this

dangell7: I agree I marked it as DNM. Still figuring out the solution.

PR #301, opened 2024-03-25
Remarks amendment

RichardAH:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/test/app/SetRemarks_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <ripple/core/ConfigSections.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Indexes.h>
+#include <ripple/protocol/TxFlags.h>
+#include <ripple/protocol/jss.h>
+#include <test/jtx.h>
+
+namespace ripple {
+namespace test {
+struct SetRemarks_test : public beast::unit_test::suite
```
RichardAH, line 30: when these tests are finished I'm probably ready to merge

src/ripple/app/tx/impl/SetRemarks.cpp
```diff
@@ -0,0 +1,451 @@
...
+        return ret;
+
+    auto& tx = ctx.tx;
+    auto& j = ctx.j;
+
+    if (tx.isFieldPresent(sfFlags) && tx.getFieldU32(sfFlags) != 0)
+    {
+        JLOG(j.warn()) << "SetRemarks: Invalid flags set.";
+        return temMALFORMED;
+    }
```
tequdev: should only allow `lsfImmutable`?

And `temINVALID_FLAG` is better than`temMALFORMED` here

RichardAH: the immutable flag is only for each individual remark object in the remarks array, it's not a transaction level flag. good point about the tem code though

tequdev: In case for tfFullyCanonicalSig is specified,
`if (ctx.tx.getFlags() & tfUniversalMask)`
would be better, no?
This PR may not be the right place to address this, but ClaimReward has the same problem.

RichardAH: Can you please PR claimreward

tequdev: https://github.com/Xahau/xahaud/pull/487

PR #308, opened 2024-05-13
Fix account_tx sluggishness

WietseWind:
Implements fix as per https://github.com/XRPLF/rippled/commit/2e9261cb

- Pre fix: ignoring index, doing full table scan. Problematic on accounts with many records / large DB (FH server).

As per Ripple Eng:

> The SQLite db is not using indexes with the previous SQL, so full table scan was going on. The fh nodes has a big table so the thread is hanging
> I found the issue but it’s the developer who found the SQLite optimizer problem 
> It only happens when the account has a lot of transactions and marker is used for paging


dangell7 (approved):

WietseWind:
Already merged into `jshooks`; - will reach DEV branch that way too.
https://github.com/Xahau/xahaud/pull/330

PR #384, opened 2024-10-24
Improve Admin command RPC Post

WietseWind:
Admin command RPC Post to URL had a 32 queue length (hardcoded) resulting in dropping TX notifications.

As this is an admin-command only, I stripped out the entire queue length check. If admin, you should know what you are doing. If your endpoint can't efficiently handle the TPS, your problem.

Also: shorter TTL for outgoing RPC HTTP calls: was 10 minutes PER REQUEST, now is 30 seconds (still too long, but 10 minutes is a guaranteed shit show if the calls keep on hanging and stack up, especially since the 32 queue length for HTTP calls is now removed).


RichardAH:
I think this is fine, the rpc subscriptions are guarded here https://github.com/Xahau/xahaud/blob/1e45d4120c687346e24fc374a849ec094167eda3/src/ripple/rpc/handlers/Subscribe.cpp#L51 and I don't think there's any other way to create them

RichardAH (approved):

WietseWind:
> I think this is fine, the rpc subscriptions are guarded here
> 
> https://github.com/Xahau/xahaud/blob/1e45d4120c687346e24fc374a849ec094167eda3/src/ripple/rpc/handlers/Subscribe.cpp#L51
> 
> and I don't think there's any other way to create them

My thoughts exactly. Thanks for taking a look.

Now running this build locally to test :)

This also finally explains why push notifications @ Xaman weren't showing for larger ledgers for some users.

PR #386, opened 2024-10-28
Fix for jshooks at macos build

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell7:
This could also be done with an else if operation. Apple has added this to clang 17 so we are just waiting. Or I'm also fine with commenting out until then

RichardAH:
yeah can we do this with preprocessor macros so we don't revert to the much slower implementation of the space ship operator if we don't need to?
```
#ifdef __APPLE__
    // macOS-specific code here
#elif
   // existing code
#endif
```

tequdev:
Since the problem with the base_unit.h file is not only in the jshooks branch but also in the dev branch, it may be better to create a separate PR?

dangell7:
> Since the problem with the base_unit.h file is not only in the jshooks branch but also in the dev branch, it may be better to create a separate PR?

Yes I would suggest that

tequdev:
#391

RichardAH:
i'll back merge dev into jshooks so I don't think this is needed anymore?

tequdev:
> i'll back merge dev into jshooks so I don't think this is needed anymore?

Only the `ripple/json/json_reader.h` include fix remains, but it can do either in this PR or directly in the jshook branch.

PR #388, opened 2024-10-29
IOUIssuerWeakTSH

### LINKED ISSUE #137: Trouble executing Weak TSH on cancel transactions
I had trouble executing weak tsh on transactions that involve removing the ledger object on execution.

On [this](https://github.com/Xahau/xahaud/tree/tsh-tests) branch if you:

1. Enable the `testCheckCancelTSH` in [this](https://github.com/Xahau/xahaud/blob/b04fe081749b191e8a5792eff2d06cdabe31c065/src/test/app/SetHookTSH_test.cpp) file and disable the current test `testOfferCreateTSH`
2. Build and run `./rippled -u ripple.app.SetHookTSH`
2. Then notice the `CHECK ID:` and `NO CHECK` std::cout logs.

I believe that its might be possible the TSH is not being called until after the check is removed (weak == after execution)

### LINKED ISSUE #262: Add weakTSH to Token Issuer
For transactions with an issued token associated, the issuer of that token should be set as at least weakTSH.

Like Payment transaction between third parties with IOU, CheckCash with IOU etc...

Perhaps need to add to this process not only the account whose balance has changed, but also the account of the issuer of the token.

https://github.com/Xahau/xahaud/blob/dev/src/ripple/app/tx/impl/Transactor.cpp#L1447-L1472

### LINKED ISSUE #297: Remit does not call the token issuer's Hook as weakTSH. (Version: 2024.3.12-release+790 )
## Issue Description
If a token is specified in the Amounts field of a Remit transaction, the Hook of that token issuer will not be invoked.

## Expected Result
The issuer of the token being sent must be weakTSH.

## Actual Result
The issuer of the token being sent is not TSH.

## Environment
xahaud 2024.3.12-release+790

## Supporting Files
https://gist.github.com/tequdev/dd19dfaacca7bd3372b646f423004a03

tequdev:
- #262
- #297
- #137

Affected to 
- CheckCreate
- CheckCash
- EscrowCancel
- EscrowCreate
- EscrowFinish
- OfferCreate
- Payment
- PaymentChannelClaim
- PaymentChannelCreate
- PaymentChannelFund
- URITokenBuy
- Remit

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -1465,8 +1465,17 @@ Transactor::addWeakTSHFromSandbox(detail::ApplyViewBase const& pv)
 
             AccountID const& lowAcc = std::get<0>(tpl);
             AccountID const& highAcc = std::get<1>(tpl);
-            STAmount const& amt = entry.second;
-            additionalWeakTSH_.emplace(amt >= beast::zero ? lowAcc : highAcc);
+            if (ctx_.view().rules().enabled(featureIOUIssuerWeakTSH))
+            {
+                additionalWeakTSH_.emplace(lowAcc);
+                additionalWeakTSH_.emplace(highAcc);
```
RichardAH: The code in which additionalWeakTSH_ is processed probably needs to be changed to ensure if an account appears twice as a result of this change that it will always use the strongest TSH status
https://github.com/Xahau/xahaud/blob/29b978bad9d06bbdf4feb0c0c0183769bd5216d3/src/ripple/app/tx/impl/Transactor.cpp#L1514

tequdev:
Need to resolve WeakTSH handling issue when transaction results in deletion of ledger object

https://github.com/Xahau/xahaud/issues/137

src/ripple/app/tx/impl/Transactor.cpp
tequdev: strongTSH is added by getTransactionalStakeHolders(), and weakTSHs are added to the end of that vector list.

Therefore, in the alreadyProcessed processing, strongTSH won't be processed after weakTSH, right?

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -131,12 +147,21 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
...
+            std::optional<STAmount> amount;
+            if (ut->isFieldPresent(sfAmount))
+                amount = ut->getFieldAmount(sfAmount);
 
             // three possible burn scenarios:
             //  the burner is the owner and issuer of the token
             //  the burner is the owner and not the issuer of the token
             //  the burner is the issuer and not the owner of the token
 
+            if (iouIssuerWeakTSH && amount)
```
RichardAH: amount.has_value() is preferred way to test optionals

RichardAH: I'm also wondering if this should be a weak tsh? This amount field on the uritoken is for a sale offer, but here the token is being destroyed. I guess it doesn't hurt but it's very very ancillary to the amount's issuer. None of their tokens change hands here.

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -363,16 +460,40 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
 
         // simple two party transactions
         case ttPAYMENT:
```
RichardAH: seems like this would be the main use case for the iouissuerweaktsh?

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -2054,7 +2065,10 @@ Transactor::operator()()
         hook::HookStateMap stateMap;
         std::vector<hook::HookResult> weakResults;
 
-        doTSH(false, stateMap, weakResults, proMeta);
+        if (!view().rules().enabled(featureIOUIssuerWeakTSH))
```
RichardAH, line 2066: is this condition supposed to be inverted?

tequdev: This PR changes getTransactionalStakeHolders() to be called only once, but since a second call is still needed when the feature is disabled, it is called here.

I will add a comment.

src/ripple/app/hook/impl/applyHook.cpp
tequdev: If using PaymentSandBox, it will be processed by `Transactor::addWeakTSHFromSandbox()`.

(ttPayment, ttOFFER_CREATE)

https://github.com/Xahau/xahaud/blob/97d1ba4e2cfc456897eca3fdeca5d64a269295e0/src/ripple/app/tx/impl/Transactor.cpp#L1480

src/ripple/app/hook/impl/applyHook.cpp
tequdev: Yes, we need consistency.

1. Only treat as TSH when there is a change in the IOU Balance

or

2. Treat as TSH when IOU is specified in the transaction (regardless of whether there is a change in the Balance) (and Balance changed)

RichardAH: Tbh it's kind of beyond a weak tsh, like "ultra weak" or ancillary or hypothetical tsh. Because the Issuer isn't a party to the transaction, isn't a stakeholder in the transaction, what we're actually saying is that they might become a party to a transaction involving this object later, and specifically in this case: that they will no longer be a party to a hypothetical future transaction involving this object later. Very ancillary indeed.

I guess my question is: does the meticulous chasing of these ancillary edges cases actually make weak tsh too expensive or less useful for its intended usecase? Should we instead add a third category of tsh for these cases and let them opt into that maybe? I suppose they can always use HookOn to ensure they're opting out, unless they're both an issuer for a currency and the issuer of uritokens.

PR #392, opened 2024-11-15
HookCanEmit

### LINKED ISSUE #365: New field in the Hook object that limits the TransactionType that can be emitted from within the Hook
Add a field that represents the same 256-bit field as HookOn and the same flags.

If the Bit of the corresponding TransactionType is high, the TT can be emitted from its Hook; if it is low, the emit will result in an error if attempted.

tequdev:
close #365

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


RichardAH:
hopefully we'll get this into the release after next

tequdev:
> hopefully we'll get this into the release after next

Due to waiting for the probability of a deterministic build, I don’t add enough tests at this time.

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -714,23 +724,9 @@ SetHook::preflight(PreflightContext const& ctx)
...
-                    << "HookSet(" << hook::log::HOOK_INVALID_FIELD << ")["
-                    << HS_ACC()
-                    << "]: Malformed transaction: SetHook sfHook contains "
-                       "invalid field.";
-                return temMALFORMED;
-            }
-        }
+        if (!ctx.rules.enabled(featureHookEmit) &&
+            hookSetObj.isFieldPresent(sfHookEmit))
+            return temMALFORMED;
```
RichardAH: temDISABLED

RichardAH:
I think HookEmit is a bit of a confusing label? Maybe HookEmittable? Or HookCanEmit

tequdev:
> hookSetObj.isFieldPresent(sfHookEmit))

@RichardAH 
Yes, at this moment I have chosen HookEmit because of its simplicity and the similarity of its sound to HookOn.
I have no objection to changing the name to something else.

sublimator:
> I think HookEmit is a bit of a confusing label? Maybe HookEmittable? Or HookCanEmit

I agree it's not totally clear even though it is nice and concise. 

There's arguments to be made for various names that prioritize brevity, for consistency with existing fields etc, but clarity is important for approachability too. I've chatted to various AIs and they'll be adamant initially that HookEmit is the best, but then can yield to various arguments. 

I was drawn to HookCanEmit initially as it's very clear. 

What about sfEmitMask ? The hook definition already has sfCreateCode as exception to any perceived Hook* naming convention.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -1395,6 +1406,16 @@ Transactor::doHookCallback(
         if (hookObj.getFieldH256(sfHookHash) != callbackHookHash)
             continue;
 
+        // default allows all transaction types
+        uint256 defaultHookEmit = UINT256_BIT[ttHOOK_SET];
+
+        uint256 hookEmit =
+            (hookObj.isFieldPresent(sfHookEmit)
+                 ? hookObj.getFieldH256(sfHookEmit)
```
sublimator: Maybe a helper function for this

tequdev: https://github.com/Xahau/xahaud/pull/392/commits/225e02afed95664d968e37c5b5a4292a9e887e06

tequdev:
`HookEmit` has been renamed `HookCanEmit`.

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -452,6 +455,13 @@ SetHook::validateHookSetEntry(SetHookCtx& ctx, STObject const& hookSetObj)
                 return false;
             }
 
+            // validate sfHookCanEmit
+            // HookCanEmit field is an optional field for backward compatibility
+            if (!hookSetObj.isFieldPresent(sfHookCanEmit))
+            {
+                // pass
```
RichardAH, line 462: What's this block for?

tequdev: I had added this block for readability to match the processing of other existing fields.

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -205,7 +205,7 @@ validateHookParams(SetHookCtx& ctx, STArray const& hookParams)
 // infer which operation the user is attempting to execute from the present and
 // absent fields
 HookSetOperation
-SetHook::inferOperation(STObject const& hookSetObj)
+SetHook::inferOperation(SetHookCtx& ctx, STObject const& hookSetObj)
```
RichardAH: why was this parameter added? it's not used?

tequdev: I used to check the temDISABLED field for the HookCanEmit field.
I'll delete it because it’s checked outside inferOperation now.

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -714,23 +724,9 @@ SetHook::preflight(PreflightContext const& ctx)
...
-                name != sfHookApiVersion && name != sfFlags)
-            {
-                JLOG(ctx.j.trace())
-                    << "HookSet(" << hook::log::HOOK_INVALID_FIELD << ")["
-                    << HS_ACC()
-                    << "]: Malformed transaction: SetHook sfHook contains "
-                       "invalid field.";
-                return temMALFORMED;
-            }
-        }
```
RichardAH: did you remove this code because it's checked by the template system already?

tequdev: Yes, Also, it is not possible to check optional fields that are not specified in the transaction within this process. (This is because all fields of the Hook object, including optional fields, are included)

I think a similar process was used in sFMemos as well.

RichardAH: It's just a whitelist of allowable fields inside the inner hookset obj.

Is there currently a unit test for this? If not can you please add one (illegal fields inside hookset obj) and ensure the preflight is still temMALFORMED in the revised code?

tequdev: Currently, as a result of the template check, this is an invalidTransaction error, not a temMALFORMED error.

The error occurs at the time of the submit command.


I can add tests to check for invalidTransaction errors, but generally such tests are not present in other test cases.

RichardAH: I think from a defensive programming standpoint there's no harm in leaving this code in? Just add sfCanEmit to the whitelist?

tequdev: added

PR #406, opened 2024-12-11
ExtendedHookState

tequdev:
- Extended the size of HookStateData, and cost one incremental reserve for every 256 bytes
- The maximum HookStateData size is tentatively set to 2048 (256*8) / Discussion needed

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1116,30 +1128,54 @@ hook::setHookState(
...
+        uint32_t newOwnerCount = computeHookStateOwnerCount(data);
+
+        uint32_t changedOwnerCount = newOwnerCount - oldOwnerCount;
+        newHookStateCount += changedOwnerCount;
+
+        if (changedOwnerCount != 0)
+        {
+            ownerCount += changedOwnerCount;
+            XRPAmount const newReserve{view.fees().accountReserve(ownerCount)};
+            if (STAmount((*sleAccount)[sfBalance]).xrp() < newReserve)
```
tequdev: TODO: Don't return `tecINSUFFICIENT_RESERVE` if the OwnerCount decreases.

tequdev: This check is not made because set_state_cache, which performs the reserve check, does not access the StateData before the change.

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -922,7 +926,7 @@ SetHook::destroyNamespace(
         sleAccount->setFieldU32(sfHookStateCount, stateCount);
 
     if (ctx.rules.enabled(fixNSDelete))
-        adjustOwnerCount(view, sleAccount, -toDelete.size(), ctx.j);
+        adjustOwnerCount(view, sleAccount, -toDeleteOwnerCount, ctx.j);
```
tequdev: TODO: needs tests that OwnerCount decreases depending on HookState size when NS deleted.

tequdev (commented):
TODO: Tests for Insufficient Reserves

RichardAH:
I think we should not add the complexity of incremental reserve increase, rather just increase the allowable size for the same existing reserve cost. I think 1 reserve for 2048 bytes is probably ok.

An alternative simpler fee model would be to nominate your account for "large hook data" whereupon all your hook states cost more to store but you have access to the larger size. I really dislike the idea of a variable length field claiming different reserves depending on its current size.

tequdev:
> I think we should not add the complexity of incremental reserve increase, rather just increase the allowable size for the same existing reserve cost. I think 1 reserve for 2048 bytes is probably ok.
> 
> An alternative simpler fee model would be to nominate your account for "large hook data" whereupon all your hook states cost more to store but you have access to the larger size. I really dislike the idea of a variable length field claiming different reserves depending on its current size.

Can you explain “large hook data” in more detail?

RichardAH:
> Can you explain “large hook data” in more detail?

Sure, so you do an account set flag which is like the equivalent of specifying large pages in your OS. All your hook states cost more but they're allowed to be bigger. It makes the computation easier. So let's say in large mode you get 2048 bytes per hook state but in computing reserve requirements for your hook state we take your HookStateCount and multiply it by 8. Once the flag is enabled it can only be disabled if HookStateCount is 0.

tequdev:
> > Can you explain “large hook data” in more detail?
> 
> Sure, so you do an account set flag which is like the equivalent of specifying large pages in your OS. All your hook states cost more but they're allowed to be bigger. It makes the computation easier. So let's say in large mode you get 2048 bytes per hook state but in computing reserve requirements for your hook state we take your HookStateCount and multiply it by 8. Once the flag is enabled it can only be disabled if HookStateCount is 0.

That's interesting.
Instead of using the flag, how about using a numeric field to represent the maximum size the account allows?

If we use the flag, whenever we need to increase the protocol's maximum allowed size, we end up having to add a new flag.

RichardAH:
> That's interesting. Instead of using the flag, how about using a numeric field to represent the maximum size the account allows?
> 
> If we use the flag, whenever we need to increase the protocol's maximum allowed size, we end up having to add a new flag.

I think that's reasonable. We could say HookStateScale or something 0,1 ...

src/ripple/app/hook/applyHook.h
```diff
@@ -32,6 +32,7 @@ class HookStateMap : public std::map<
                          std::tuple<
                              int64_t,  // remaining available ownercount
                              int64_t,  // total namespace count
+                             int16_t,  // hook state scale
```
RichardAH: I don't mind if you use int16_t or uint16_t or uint32_t but keep it consistent with the function calls? like `maxHookStateDataSize(uint32_t hookStateScale)` same for the sf type

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -921,7 +927,7 @@ SetHook::destroyNamespace(
         view.erase(sleItem);
     }
 
-    uint32_t stateCount = oldStateCount - toDelete.size();
+    uint32_t stateCount = oldStateCount - toDeleteStateCount;
```
RichardAH: was there a reason for changing this? I assume it doesn't change the outcome

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -937,7 +943,7 @@ SetHook::destroyNamespace(
         sleAccount->setFieldU32(sfHookStateCount, stateCount);
 
     if (ctx.rules.enabled(fixNSDelete))
-        adjustOwnerCount(view, sleAccount, -toDelete.size(), ctx.j);
+        adjustOwnerCount(view, sleAccount, -toDeleteStateCount * scale, ctx.j);
```
RichardAH: we should do a sanity check on this value before passing it to make sure it is reasonable and hasn't overflowed

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1517,19 +1528,22 @@ set_state_cache(
         stateMap.modified_entry_count++;
 
         stateMap[acc] = {
-            availableForReserves - 1,
+            availableForReserves - hookStateScale,
```
RichardAH, line 1540: we need defensive sanity checks on math like this

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1547,7 +1561,7 @@ set_state_cache(
                 namespaceCount++;
             }
 
-            availableForReserves--;
+            availableForReserves -= hookStateScale;
```
RichardAH, line 1577: overflow sanity check needed

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1559,11 +1573,12 @@ set_state_cache(
     auto& stateMapNs = stateMapAcc[ns];
     if (stateMapNs.find(key) == stateMapNs.end())
     {
+        // new State Key
         if (modified)
         {
             if (!canReserveNew)
                 return RESERVE_INSUFFICIENT;
-            availableForReserves--;
+            availableForReserves -= hookStateScale;
```
RichardAH, line 1599: here too

src/ripple/protocol/impl/TER.cpp
```diff
@@ -94,6 +94,8 @@ transResults()
         MAKE_ERROR(tecINSUF_RESERVE_SELLER,          "The seller of an object has insufficient reserves, and thus cannot complete the sale."),
         MAKE_ERROR(tecIMMUTABLE,                     "The remark is marked immutable on the object, and therefore cannot be updated."),
         MAKE_ERROR(tecTOO_MANY_REMARKS,              "The number of remarks on the object would exceed the limit of 32."),
+        MAKE_ERROR(tecHAS_HOOK_STATE,                "The account has hook state."),
```
RichardAH: maybe expand the explanation to "The account has hook state. Delete all existing state before changing the scale."

src/ripple/app/hook/impl/applyHook.cpp
tequdev: 91f683a9994e02c29f85cd3aede807bb6ebee93f

src/ripple/app/hook/impl/applyHook.cpp
tequdev: 91f683a9994e02c29f85cd3aede807bb6ebee93f

src/ripple/app/hook/impl/applyHook.cpp
tequdev: 91f683a9994e02c29f85cd3aede807bb6ebee93f

src/ripple/app/tx/impl/SetHook.cpp
tequdev: https://github.com/Xahau/xahaud/commit/a75daaea0249c1fabd773c4ef1ee08929cd9d3a3

src/ripple/app/tx/impl/SetHook.cpp
tequdev: https://github.com/Xahau/xahaud/commit/a75daaea0249c1fabd773c4ef1ee08929cd9d3a3

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -249,6 +261,20 @@ SetAccount::preclaim(PreclaimContext const& ctx)
...
+    // HookStateScale
+    if (ctx.tx.isFieldPresent(sfHookStateScale))
+    {
+        uint16_t const newScale = ctx.tx.getFieldU16(sfHookStateScale);
+        uint16_t const currentScale = sle->getFieldU16(sfHookStateScale);
+        uint32_t const stateCount = sle->getFieldU32(sfHookStateCount);
+        if (stateCount > 0 && newScale < currentScale)
+        {
+            JLOG(ctx.j.trace())
+                << "Cannot decrease HookStateScale if state count is not zero.";
```
RichardAH, line 274: shouldn't be able to set/change it at all unless hook state count is zero

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -629,6 +655,44 @@ SetAccount::doApply()
...
+            sle->makeFieldAbsent(sfHookStateScale);
+        }
+        else
+        {
+            // increase OwnerCount
+            uint32_t const stateCount = sle->getFieldU32(sfHookStateCount);
+            uint32_t const oldOwnerCount = sle->getFieldU32(sfOwnerCount);
+
+            uint32_t const newOwnerCount =
+                oldOwnerCount - (oldScale * stateCount) + (scale * stateCount);
```
RichardAH: sanity check newOwnerCount > oldOwnerCount

src/ripple/protocol/impl/TER.cpp
```diff
@@ -94,6 +94,8 @@ transResults()
         MAKE_ERROR(tecINSUF_RESERVE_SELLER,          "The seller of an object has insufficient reserves, and thus cannot complete the sale."),
         MAKE_ERROR(tecIMMUTABLE,                     "The remark is marked immutable on the object, and therefore cannot be updated."),
         MAKE_ERROR(tecTOO_MANY_REMARKS,              "The number of remarks on the object would exceed the limit of 32."),
+        MAKE_ERROR(tecHAS_HOOK_STATE,                "The account has hook state. Delete all existing state first."),
```
RichardAH: Change comment to something like: Delete all hook state before reducing scale.

tequdev: Since the number of TER codes is limited, we used a generic error message. However, we will change it as you pointed out, as it can be modified later when this code is used for other purposes.

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -184,6 +184,18 @@ SetAccount::preflight(PreflightContext const& ctx)
...
 
+    // HookStateScale
+    if (tx.isFieldPresent(sfHookStateScale))
+    {
+        if (!ctx.rules.enabled(featureExtendedHookState))
+            return temMALFORMED;
+
+        uint16_t scale = tx.getFieldU16(sfHookStateScale);
+        if (scale == 0 ||
+            scale > 16)  // Min: 1, Max: 16 (256 * 16 = 4096 bytes)
```
sublimator: extract a constant for this ?

src/ripple/app/hook/Enum.h
```diff
@@ -44,9 +46,14 @@ maxHookParameterValueSize(void)
...
-maxHookStateDataSize(void)
+maxHookStateDataSize(uint16_t hookStateScale)
 {
-    return 256U;
+    if (hookStateScale == 0)
+    {
+        // should not happen, but just in case
+        return 256U;
+    }
+    return 256U * hookStateScale;
```
sublimator, line 67: we could add an extra defensive check here with a constant, I guess

src/ripple/app/tx/impl/SetAccount.cpp
sublimator: This does make the feature hard to test out without trapping yourself though, right? Buyer beware?

src/test/rpc/AccountSet_test.cpp
```diff
@@ -574,6 +576,93 @@ class AccountSet_test : public beast::unit_test::suite
...
+        jt[sfHookStateScale.fieldName] = 4;
+        env(jt, ter(tecHAS_HOOK_STATE));
+        // increase
+        jt[sfHookStateScale.fieldName] = 6;
+        env(jt);
+        BEAST_EXPECT(env.le(alice)->getFieldU16(sfHookStateScale) == 6);
+        BEAST_EXPECT(env.le(alice)->getFieldU32(sfHookStateCount) == 10);
+        BEAST_EXPECT(env.le(alice)->getFieldU32(sfOwnerCount) == 110);
+    }
+
```
sublimator, line 676: 1. Decrease WITHOUT state (the escape hatch)

  applyCount(5, 0, 50);  // ← stateCount=0 (all state deleted)
  jt[sfHookStateScale.fieldName] = 2;
  env(jt);  // ← Should succeed! This is how you escape high scale
  BEAST_EXPECT(env.le(alice)->getFieldU16(sfHookStateScale) == 2);

  This is important because it tests the only way to reduce scale - by deleting all state first. The spec calls this the "Scale Commitment Trap" escape route.

  2. Insufficient reserves on increase

  // Alice has exactly enough for current reserves, but not for 16x increase
  applyCount(1, 100, 200);  // 100 entries at scale=1
  // Manually set balance to just cover current reserves
  jt[sfHookStateScale.fieldName] = 16;  // Needs 1600 reserves (100×16)
  env(jt, ter(tecINSUFFICIENT_RESERVE));  // Should fail - can't afford it

src/ripple/app/hook/Enum.h
tequdev: like this?
```cpp
inline uint32_t
maxHookStateDataSize(uint16_t hookStateScale)
{
    if (hookStateScale == 0)
    {
        // should not happen, but just in case
        return 256U;
    }
    if (hookStateScale > maxHookStateScale())
    {
        // should not happen, but just in case
        return 256 * maxHookStateScale();
    }
    return 256U * hookStateScale;
}
```

sublimator: I think assert/throw "fail fast" because it's a violated invariant indicating something went awry ?

tequdev: I think it's too much to add that kind of processing here.

sublimator: Well, we can do "return 256U * hookStateScale;" if it's just a simple calculation, but if you want to handle strange cases (0 || > max_scale) this way, it means you're explicitly washing over bad state. If it should not happen, assert that it doesn't happen. This is some weird middle ground, no?

sublimator (dismissed):

PR #412, opened 2024-12-16
Fix Error handling on build action

### LINKED ISSUE #409: Github Action shows success for job even though build has error.
https://github.com/Xahau/xahaud/actions/runs/12267969413/job/34229365811?pr=318
スクリーンショット 2024-12-16 14 42 56

tequdev:
close #409 

Example of failure in build job.
https://github.com/Xahau/xahaud/actions/runs/12348238153/job/34456701369

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/build-core.sh
+++ b/build-core.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 echo "START INSIDE CONTAINER - CORE"
 
 echo "-- BUILD CORES:       $3"
--- a/build-full.sh
+++ b/build-full.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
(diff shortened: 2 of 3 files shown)


build-core.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
sublimator, line 3: LGTM, and a huge step in the right direction, but also propose:

```
#!/bin/bash -u
# We use set -e and bash with -u to bail on first non zero exit code of any
# processes launched or upon any unbound variable.
# We use set -x to print commands before running them to help with
# debugging.
set -ex
```

tequdev: I will add these after the current CI is finished.

PR #415, opened 2024-12-24
Fix HookResult(ExitType) when accept() is not called

### LINKED ISSUE #354: Show rollback message or error message in case of missing the `accept` hook api
## Steps to Reproduce
```cpp
#include "hookapi.h"

int64_t hook(uint32_t reserved) {
  _g(1, 1);
  return 0;
}
```

## Actual Result

```
HookInfo[rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn-rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn]: creating wasm instance
HookInfo[rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn-rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn]: ApiVersion: 0 ROLLBACK RS: '' RC: -1
```

tequdev:
close #354

Change ExitType to UNSET if `accept()` hook api is not called

Fixed as an amendment because of the effect on meta.HookExecution.HookResult

### Type of Change

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/ripple/app/hook/impl/applyHook.cpp
+++ b/src/ripple/app/hook/impl/applyHook.cpp
@@ -1217,9 +1217,10 @@ hook::apply(
              .hookParamOverrides = hookParamOverrides,
              .hookParams = hookParams,
              .hookSkips = {},
-             .exitType =
-                 hook_api::ExitType::ROLLBACK,  // default is to rollback unless
-                                                // hook calls accept()
+             .exitType = applyCtx.view().rules().enabled(fixXahauV1)
+                 ? hook_api::ExitType::UNSET
+                 : hook_api::ExitType::ROLLBACK,  // default is to rollback
+                                                  // unless hook calls accept()
              .exitReason = std::string(""),
              .exitCode = -1,
              .hasCallback = hasCallback,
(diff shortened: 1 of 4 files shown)


RichardAH:
The default is set to rollback on purpose... It's a security measure. A badly programmed hook might reach the end of the hook function without calling accept() or rollback() via some user input trickery. In these cases rollback is preferable because it's much less likely to result in a hack/loss of funds.

tequdev:
The mapping of ExitType::ROLLBACK->tecHOOK_REJECTED condition seems to be determined by `hookResult.exitType ! = hook_api::ExitType::ACCEPT`

https://github.com/Xahau/xahaud/blob/effc920b7e8ed3ccdbe48de1121cd32e3a87ad7b/src/ripple/app/tx/impl/Transactor.cpp#L1268-L1286

But if there is another way to resolve #354, that would be fine.

RichardAH:
Ok I've thought about it a while and come around to your position, merging thanks :)

PR #426, opened 2025-01-27
ServiceFees

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/app/tx/impl/Transactor.cpp
+++ b/src/ripple/app/tx/impl/Transactor.cpp
@@ -2040,7 +2040,7 @@ Transactor::operator()()
             result = tecOVERSIZE;
     }
 
-    if (view().rules().enabled(featureServiceFee) && applied &&
+    if (applied && view().rules().enabled(featureServiceFee) &&
         ctx_.tx.isFieldPresent(sfServiceFee))
         do
         {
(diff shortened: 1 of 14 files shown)


src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -2040,14 +2040,233 @@ Transactor::operator()()
             result = tecOVERSIZE;
     }
 
+    if (applied && view().rules().enabled(featureServiceFee) &&
```
tequdev: If the ServiceFee field is included when the ServiceFee amendment is not enabled, shouldn't the transaction be blocked (preflight level)?

tequdev (changes requested):

src/ripple/app/tx/impl/Transactor.cpp
RichardAH: yes good catch

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -1894,6 +1900,214 @@ Transactor::operator()()
...
+                    << "skipping self service-fee on " << src << ".";
+                break;
+            }
+
+            if (amt <= beast::zero)
+            {
+                JLOG(j_.info())
+                    << "skipping non-positive service-fee from " << src << ".";
+                break;
+            }
```
tequdev: The only reason why ServiceFee was not used in the transaction should be due to information in the on-ledger data, such as insufficient balance, noAuth, etc. 
Those should be a tem error.

dangell7: updated

PR #432, opened 2025-02-01
Add Conan Building For Development

### LINKED ISSUE #402: Support Conan
Currently I'm using the conan-strict branch by @dangell7 as a base during development.
It would be great if this could be used in the dev branch as well.

https://github.com/Xahau/xahaud/tree/conan-strict

However, Boost 1.86 used in the current dev branch causes build errors, so I have downgraded to 1.83 in my environment.
https://github.com/tequdev/xahaud/commit/ec9c62541ecf1a559bb96b5e12f30c6a0c7c2671

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


tequdev:
It is currently working fine in my local build.

src/test/server/ServerStatus_test.cpp
```diff
@@ -677,7 +677,7 @@ class ServerStatus_test : public beast::unit_test::suite,
             std::string(resp["Upgrade"]) == "websocket");
         BEAST_EXPECT(
             resp.find("Connection") != resp.end() &&
-            std::string(resp["Connection"]) == "Upgrade");
+            std::string(resp["Connection"]) == "upgrade");
```
dangell7: @tequdev on the release builder its "Upgrade" on the clang is "upgrade" lol

tequdev:
Both build and test are successful locally.

tequdev (commented):
We should make sure that the workflow is successful after clearing the cache once before marge.

Builds/CMake/conan/Boost.cmake
```diff
@@ -0,0 +1,51 @@
+find_package(Boost 1.83 REQUIRED
```
tequdev, line 1: Boost::iostreams has been added by Catalog, so Boost.cmake for Conan also needs to be updated.

https://github.com/Xahau/xahaud/pull/443/files

tequdev:
Considering rippled sync, I think it is valuable to keep the workflow YML file names consistent with the rippled file names.

dangell7:
> Considering rippled sync, I think it is valuable to keep the workflow YML file names consistent with the rippled file names.

I would agree however @sublimator did some fancy stuff and were also not using hosted runners. So what if we just keep those old files and then disabled the workflow in github so theres not merge conflicts?

sublimator:
Ja, I made some of the files Xahau namespaced so it's easy to use our now
diverged GH runners without fighting conflicts later. Ripple seems to have
some serious resources dedicated to their self hosted runners. It also
makes it easier to see at a glance what is surplus/different to rippled.
That said, I am on phone atm and not sure exactly what is being discussed

src/test/app/SetHook_test.cpp
```diff
@@ -1110,7 +1110,6 @@ class SetHook_test : public beast::unit_test::suite
         env.fund(XRP(10000000), bob);
 
         auto const preHookCount = (*env.le(alice))[sfHookStateCount];
-        auto const preOwnerCount = (*env.le(alice))[sfOwnerCount];
```
sublimator: I'm less religious about keeping PRs totally focused, but curious why this is happening in this PR?

dangell7: Because at the time we were running the PR against the ripple workflows which are very strict in the builds, forcing warnings into errors. I think that was the reason or I just went on a warning fix tyrant.

sublimator: Figured might be something like that :)

sublimator (dismissed):

src/ripple/app/rdb/backend/FlatmapDatabase.h
```diff
@@ -67,7 +65,7 @@ class FlatmapDatabase : public SQLiteDatabase
 
 public:
     FlatmapDatabase(Application& app, Config const& config, JobQueue& jobQueue)
-        : app_(app), config_(config), jobQueue_(jobQueue)
```
RichardAH: why are these types of changes in the conan pr?

dangell7: When we merged the conan code it, it came with github workflows. Those workflows held the build to a stricter standard and the unused variables were flagged and thrown. 

Since we switched to a different build, less strict I can remove the code that causes warnings, or fix the in another PR first?

sublimator:
image

These seem to be required checks that are no longer in the github workflows ?

src/ripple/app/rdb/backend/FlatmapDatabase.h
sublimator: Separate PR merged before this sounds good!?

dangell7:
> image
> These seem to be required checks that are no longer in the github workflows ?

Yeah I changed it into one job with many steps. I will update the required deployments/actions once its merged

src/ripple/app/rdb/backend/FlatmapDatabase.h
RichardAH: yeah I just don't want code changes in a build system pr... it's too much mixing. if we need to revert build system pr we shouldn't be rolling back code changes at the same time, it's messy

RichardAH:
Is this ready now? should something else be merged first?

tequdev:
> We should make sure that the workflow is successful after clearing the cache once before marge.

I think this is needed.

we can delete cache from here.
https://github.com/Xahau/xahaud/actions/caches

RichardAH:
> > We should make sure that the workflow is successful after clearing the cache once before marge.
> 
> I think this is needed.
> 
> we can delete cache from here. https://github.com/Xahau/xahaud/actions/caches

just deleted and added a nothing commit to force a rebuild

sublimator:
Woooo

Issue #420, opened 2025-01-10
Building ServerDefinitions.cpp takes a very long time.


tequdev:
## Issue Description
As the title says

## A diff build where only ServerDefinitions.cpp was changed.

(The build uses Conan, a non-standard build of this repository.)

### Currently
`cmake --build . --target rippled --parallel -j$(sysctl -n hw.logicalcpu)  1458.46s user 122.93s system 81% cpu 32:19.25 total`

### Comment out the code that uses magic_enum:
`cmake --build . --target rippled --parallel -j$(sysctl -n hw.logicalcpu)  15.72s user 3.86s system 127% cpu 15.325 total`

### Environment

macos 15.2 / M1 Max

mvadari:
The rippled implementation of `server_definitions` doesn't use `magic_enum`, so the xahaud implementation could borrow that modification.

tequdev:
## Differences in the output of server_definitions for rippled and xahaud

### LEDGER_ENTRY_TYPES

- xahaud
    - from LedgerFormats.h LedgerEntryType
    - including deprecated ones (eg. ltCONTRACT, ltGENERATOR_MAP
- rippled
    - from LedgerFormats.cpp LedgerFormats()
    - not including deprecated ones

### TRANSACTION_FLAGS, TRANSACTION_FLAGS_INDICES

- xahaud: exists
- rippled: not exists

### TRANSACTION_RESULTS

- xahaud
    - from TER.h TELcodes, TEMcodes, TEFcodes, TERcodes, TEScodes, TECcodes 
    - including reserved ones(eg. AMM, XCHAINS)

- rippled
    - from TER.cpp transResults()
    - only including ones defined by MAKE_ERROR macro

### TRANSACTION_TYPES

**TRANSACTION_TYPES can have different keys even if they are the same transaction type.**

- xahaud
    - from TxFormats.h TxType
    - including deprecated ones)
    - **tt-based transaction type name** 
        - ttAmendment -> "Amendment"
        - ttFee -> "Fee"
- rippled
    - from TxFormats.cpp TxFormats()
    - only including ones defined by add())
    - **jss-based transaction type name**
        - jss::EnableAmendment -> "EnableAmendment"
        - jss::Fee -> "Fee"

### native_currency_code
- xahaud: exists
- rippled: not exists (but XRPL Cluster has)

tequdev:
#436

PR #436, opened 2025-02-12
Fixed not to use a large fixed range in the magic_enum.

tequdev:

Changed to specify the suitable range for each Enum in enum_range.
 
 #420
 
`cmake --build . --target rippled --parallel -j$(sysctl -n hw.logicalcpu)  25.63s user 7.30s system 84% cpu 39.070 total`

The build time for ServerDefinitions.cpp was r**educed by 98%** (from 1460s to 25s) on a macOS environment.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/rpc/handlers/ServerDefinitions.cpp
+++ b/src/ripple/rpc/handlers/ServerDefinitions.cpp
@@ -36,12 +36,12 @@
 #include <magic/magic_enum.h>
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


sublimator:
Without this patch clang 16 will stall on unity file 16 which is containing the definitions.

sublimator:
@dangell7 

Does this look good for dev? 
It's definitely helping the build times when using the latest clang 16

PR #443, opened 2025-02-16
Catalogue

RichardAH:
```json
{
  "command":"catalogue_create",
  "output_file":"/root/cat.12650000.12950000.hashed",
  "min_ledger": 12650000,
  "max_ledger":12950000,
  "compression_level":9
}
```
```json
{
  "command":"catalogue_load",
  "input_file":"/root/cat.3"
}
```

```json
{
  "command":"catalogue_status"
}
```

Catalogue files are history storage files that store a base ledger plus the deltas of a number of subsequent ledgers in a single file under solid zlib compression.

File integrity can be checked with this tool:
https://github.com/RichardAH/catalogue-file-validator


sublimator:
Wait, are you making history packs?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1528 @@
...
+
+            // Set the node type
+            header.type = convertNodeType(node.getType());
+
+            header.isRoot = isRoot ? 1 : 0;
+            header.deltaType = deltaType;
+            header.hash = node.getHash().as_uint256();
+
+            // For inner nodes, store the nodeID; for leaf nodes, store the key
+            if (node.isInner())
```
sublimator: Why store the inner nodes though?
You can rebuild them easily enough

sublimator: Loading rebuild speed/efficiency or something ?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+        // Create a ledger object
+        std::shared_ptr<Ledger> ledger;
+
+        if (info.seq == header.min_ledger)
+        {
+            // Base ledger - create a fresh one
+            ledger = std::make_shared<Ledger>(
+                info.seq,
+                context.app.timeKeeper().closeTime(),
```
sublimator: ?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,498 @@
...
+        return rpcError(rpcINVALID_PARAMS, "min_ledger must be <= max_ledger");
+
+    std::vector<std::shared_ptr<Ledger const>> ledgers;
+    ledgers.reserve(max_ledger - min_ledger + 1);
+
+    // Grab all ledgers of interest
+    for (auto i = min_ledger; i <= max_ledger; ++i)
+    {
+        std::shared_ptr<Ledger const> ptr;
+        auto status = RPC::getLedger(ptr, i, context);
```
sublimator: Should it return an error if the ledger is not validated yet?

src/ripple/rpc/handlers/Catalogue.cpp
sublimator: Just placeholder closeTime ?

RichardAH:
ready for a new review I think

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1139 @@
...
+        if (!outputLedger(currLedger, prevLedger->stateMap()))
+            return rpcError(
+                rpcINTERNAL, "Error occurred while processing ledgers");
+
+        ledgers_written++;
+
+        // Periodically flush the stream to prevent excessive memory usage in
+        // compression buffers Flush every 1000 ledgers as a reasonable balance
+        // between compression efficiency and memory usage
+        if (ledgers_written % 1000 == 0)
```
sublimator: Could this be every n bytes? 
And check after the first ledger full map?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1139 @@
...
+    else
+    {
+        JLOG(context.j.info())
+            << "No compression (level 0), using direct output";
+    }
+    compStream->push(boost::ref(outfile));
+
+    // Process ledgers with local processor implementation
+    auto writeToFile = [&compStream, &context](const void* data, size_t size) {
+        compStream->write(reinterpret_cast<const char*>(data), size);
```
sublimator, line 477: Could track bytes written from here

src/ripple/rpc/handlers/Catalogue.cpp
sublimator: Given "How long is a piece of ledger string?"

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1139 @@
...
+    // Create a buffer for reading
+    std::vector<char> buffer(64 * 1024);  // 64K buffer
+
+    // Read and process the header portion
+    hashFile.read(buffer.data(), sizeof(CATLHeader));
+    if (hashFile.gcount() != sizeof(CATLHeader))
+        return rpcError(rpcINTERNAL, "failed to read header for hashing");
+
+    // Zero out the hash portion in the buffer for hash calculation
+    std::fill(
```
sublimator, line 656: This is kind of "belts and suspenders" mode, I guess, because the hash is already zeroed and the file only just written:

```c++
struct CATLHeader
{
    uint32_t magic = CATL;
    uint32_t min_ledger;
    uint32_t max_ledger;
    uint16_t version;
    uint16_t network_id;
    uint64_t filesize = 0;  // Total size of the file including header
    std::array<uint8_t, 64> hash = {};  // SHA-512 hash, initially set to zeros
};
```

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1139 @@
...
+        // Store the ledger
+        prevLedger = ledger;
+        ledgersLoaded++;
+    }
+
+    decompStream->reset();
+    infile.close();
+
+    // Update ledger range in ledger master
+    context.app.getLedgerMaster().setLedgerRangePresent(
```
sublimator:
```C++
        context.app.getLedgerMaster().setLedgerRangePresent(
            header.min_ledger, info.seq, true);

        // Store the ledger
        prevLedger = ledger;
        ledgersLoaded++;
    }

    decompStream->reset();
    infile.close();

    // Update ledger range in ledger master
    context.app.getLedgerMaster().setLedgerRangePresent(
        header.min_ledger, header.max_ledger, true);
```

The last one seems redundant if it's already called as each ledger is loaded ? Actually, if you look at the loop, there's only one break rather than an early return, and in that case, it assumes success? 
And marks the full header's ledger range?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -0,0 +1,1139 @@
...
+                sizeof(closeTimeResolution)) ||
+            !decompStream->read(
+                reinterpret_cast<char*>(&closeTime), sizeof(closeTime)) ||
+            !decompStream->read(
+                reinterpret_cast<char*>(&parentCloseTime),
+                sizeof(parentCloseTime)))
+        {
+            JLOG(context.j.warn()) << "Catalogue load expected but could not "
+                                      "read the next ledger header.";
+            break;
```
sublimator: This should jump to error handling somehow, or set an error flag ?
Otherwise, per comment below, it will errantly setLedgerRangePresent for the full header range

src/ripple/rpc/handlers/Catalogue.cpp
RichardAH: Good points, I might move it to the shamap function

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -1242,4 +1242,399 @@ SHAMap::invariants() const
...
 }
 
+template <typename StreamType>
+std::size_t
+SHAMap::serializeToStream(
+    StreamType& stream,
+    std::optional<std::reference_wrapper<const SHAMap>> baseSHAMap) const
+{
+    // Static map to track bytes written to streams
+    static std::mutex streamMapMutex;
```
sublimator, line 1252: **Gemini's Comment:**

Why not make a boost stream filter (`$boost-streams-thing`) that handles all this?

The current static map/mutex (`streamMapMutex`, `streamBytesWritten`) for tracking bytes written across streams seems quite complex and introduces global state managed by stream pointer keys, which might be fragile.

Using a custom Boost.Iostreams filter would encapsulate the byte counting and threshold flushing logic cleanly. The calling code (`$usage-example`) would simply push the filter onto the stream, and `SHAMap::serializeToStream` (`$impl-example`) would become simpler, just writing to the stream without needing internal knowledge of the flushing strategy, eliminating the need for the static variables and mutex for this purpose.

**Example `$boost-streams-thing` (Custom Filter Implementation):**
```c++
#include <boost/iostreams/concepts.hpp> // sink
#include <boost/iostreams/operations.hpp> // write, flush
#include <cstddef> // std::size_t

// Define a Threshold for Flushing (e.g., 256 MB)
constexpr std::size_t DEFAULT_FLUSH_THRESHOLD = 256 * 1024 * 1024;

struct ByteCountingFlushFilter : boost::iostreams::output_filter
{
    explicit ByteCountingFlushFilter(
        std::size_t threshold = DEFAULT_FLUSH_THRESHOLD)
        : threshold_(threshold), bytes_since_flush_(0) {}

    template<typename Sink>
    bool put(Sink& /*sink*/, char /*c*/)
    {
        // Prefer write for efficiency
        return false;
    }

    template<typename Sink>
    std::streamsize write(Sink& sink, const char* s, std::streamsize n)
    {
        std::streamsize bytes_written = boost::iostreams::write(sink, s, n);
        if (bytes_written > 0)
        {
            bytes_since_flush_ += static_cast<std::size_t>(bytes_written);
            if (bytes_since_flush_ >= threshold_)
            {
                boost::iostreams::flush(sink);
                bytes_since_flush_ = 0;
            }
        }
        return bytes_written;
    }

    template<typename Sink>
    void close(Sink& sink)
    {
         if (bytes_since_flush_ > 0)
         {
             boost::iostreams::flush(sink);
             bytes_since_flush_ = 0;
         }
    }

private:
    std::size_t threshold_;
    std::size_t bytes_since_flush_;
};
```

**Example `$usage-example` (How `doCatalogueCreate` uses it):**
```c++
    // ... inside doCatalogueCreate ...
    std::ofstream outfile(filepath.c_str(), std::ios::out | std::ios::binary);
    // ... error check outfile ...

    auto compStream = std::make_unique<boost::iostreams::filtering_ostream>();

    // *** ADD THE NEW FILTER HERE ***
    // The calling code adds the filter which calls flush after threshold
    compStream->push(ByteCountingFlushFilter{}); // Uses default threshold

    if (compressionLevel > 0)
    {
        // Add compression filter *after* the flushing filter
        boost::iostreams::zlib_params params((int)compressionLevel);
        // ... set params ...
        compStream->push(boost::iostreams::zlib_compressor(params));
    }

    // Finally, push the actual file sink
    compStream->push(boost::ref(outfile));

    // --- The SHAMap::serializeToStream function is now simpler ---

    // Example call within outputLedger lambda:
    // size_t stateNodesWritten =
    //     ledger->stateMap().serializeToStream(*compStream, prevStateMap);
    // size_t txNodesWritten =
    //     ledger->txMap().serializeToStream(*compStream);
    // ... rest of doCatalogueCreate ...

    // Final flush in doCatalogueCreate is still good practice
    compStream->flush();
    compStream->reset();
    outfile.flush();
    outfile.close();
    // ...
```

**Example `$impl-example` (Simplified `SHAMap::serializeToStream`):**
*(Conceptual - showing removal of flushing logic)*
```c++
template <typename StreamType>
std::size_t
SHAMap::serializeToStream(
    StreamType& stream, // This is now potentially the filtering_ostream
    std::optional<std::reference_wrapper<const SHAMap>> baseSHAMap) const
{
    // NO static map, NO mutex, NO local byte counter for flushing needed here

    std::unordered_set<SHAMapHash, beast::uhash<>> writtenNodes;
    if (!root_)
        return 0;
    std::size_t nodeCount = 0;

    auto serializeLeaf = [&stream](SHAMapLeafNode const& node) -> bool {
        // simplified - just write
        auto t = node.getType();
        stream.write(reinterpret_cast<char const*>(&t), 1);
        auto const key = node.peekItem()->key();
        stream.write(reinterpret_cast<char const*>(key.data()), 32);
        auto data = node.peekItem()->slice();
        uint32_t size = data.size();
        stream.write(reinterpret_cast<char const*>(&size), 4);
        stream.write(reinterpret_cast<char const*>(data.data()), size);
        return !stream.fail();
    };

    auto serializeRemovedLeaf = [&stream](uint256 const& key) -> bool {
        // simplified - just write
        auto t = SHAMapNodeType::tnREMOVE;
        stream.write(reinterpret_cast<char const*>(&t), 1);
        stream.write(reinterpret_cast<char const*>(key.data()), 32);
        return !stream.fail();
    };

    // ... (Delta logic using the simplified serialize* lambdas) ...

    // ... (Full walk logic using the simplified serialize* lambdas) ...

    // Write terminal marker (simplified - just write)
    auto t = SHAMapNodeType::tnTERMINAL;
    stream.write(reinterpret_cast<char const*>(&t), 1);

    // NO final flush check here, caller handles final flush

    return nodeCount;
}

```
---

sublimator: Dunno if that actually works, but it seems cleaner :)

RichardAH: Because it makes the function I've intentionally templated type specific. The streams might be pointer fragile but if they are then they just don't get automatically flushed, the caller will flush eventually or they will flush when their internal buffers fill.

PR #440, opened 2025-02-13
Add GitHub Actions workflow for CHooks testing source

tequdev:
## High Level Overview of Change

- Ensure build_test_hooks.sh works on macOS environments
- Maintain cross-platform compatibility between Linux and macOS

Note: requires 'ggrep' to be installed via Homebrew on macOS

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Tests (you added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation update
- [ ] Chore (no impact to binary, e.g. `.gitignore`, formatting, dropping support for older tooling)
- [ ] Release

### API Impact

- [ ] Public API: New feature (new methods and/or new fields)
- [ ] Public API: Breaking change (in general, breaking changes should only impact the next api_version)
- [ ] `libxrpl` change (any change that may affect `libxrpl` or dependents of `libxrpl`)
- [ ] Peer protocol change (must be backward compatible or bump the peer protocol version)

PR #449, opened 2025-02-25
Add GitHub Actions workflow for JSHooks testing source

tequdev:

- Create a new GitHub Actions workflow for testing JSHooks sources
- Update script(SetJSHook_wasm.h) to run can be any path
    - need to place qjsc in src/test/app.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


.github/workflows/js_hooks_test.yml
```diff
@@ -0,0 +1,110 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential clang-format bc
```
sublimator, line 24: Maybe we should pin a specific version of clang-format in some environment variable that be easily changed. 

I noticed the other day that clang-format@19 would not work for some files.

src/test/app/build_test_jshooks.sh
```diff
@@ -15,18 +39,18 @@ namespace ripple {
 namespace test {
 std::map<std::string, std::vector<uint8_t>> jswasm = {' > $OUTPUT_FILE
 COUNTER="0"
-cat SetJSHook_test.cpp | tr '\n' '\f' |
-        ggrep -Po 'R"\[test\.hook\](.*?)\[test\.hook\]"' | 
-        gsed -E 's/R"\[test\.hook\]\(//g' | 
-        gsed -E 's/\)\[test\.hook\]"[\f \t]*/\/*end*\//g' | 
+cat "$INPUT_FILE" | tr '\n' '\f' |
+        grep -Po 'R"\[test\.hook\](.*?)\[test\.hook\]"' |
```
sublimator, line 43: Why not just use ggrep ? 
If it's targeting gnu grep/sed specifically ?

> brew install gnu-sed grep

This installs the gsed/ggrep commands, no?

src/test/app/build_test_jshooks.sh
```diff
@@ -48,3 +72,4 @@ echo '};
 }
 }
 #endif' >> $OUTPUT_FILE
+clang-format -i $OUTPUT_FILE
```
sublimator, line 75: Good idea!

src/test/app/build_test_jshooks.sh
```diff
@@ -1,10 +1,34 @@
...
+# Prerequisites:
+# - quickjs:
+#   $git clone https://github.com/RichardAH/quickjslite.git
+#   $cd quickjslite
+#   $make
+#   copy `qjsc` to the src/test/app of the repo
+#
+# - clang-format:
+#   Ubuntu: $sudo apt-get install clang-format
+#   macOS: $brew install clang-format
```
sublimator: brew install r-lib/taps/clang-format@10    

To match the CI clang-format I had to use that. 19 wasn't working in every case. 

related to: 
https://github.com/Xahau/xahaud/pull/449/files#r1972938606

sublimator (changes requested):
Other than the ggrep -> grep and gsed -> sed changes, this looks good to me. I think I'd rather leave the macos defaults as they are for other scripts?

src/test/app/build_test_jshooks.sh
tequdev: This script should be usable in both macos and ubuntu environments.
Therefore, scripts that are only usable in macos (gsed/ggrep) should not be used here.


you can use gsed/ggrep as sed/grep by add PATH.
```
add path: PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH"
```

sublimator: Oh, I didn't realize that ggrep and gsed weren't available on Ubuntu

It's been an embarrassingly long time since I've used linux/ubuntu

sublimator (approved):

.github/workflows/js_hooks_test.yml
tequdev: It seems that clang-format-10 is not provided in ubuntu-latest (ubuntu 22.04).

We can install clang-format-10 in ubuntu 20.04, but errors will occur around qjsc.

At the moment, there is no effect on SetJSHook_wasm.h, so I will leave it as it is.
If you could fix it, I would be very grateful.

sublimator:
We better double check me re: cf 10
I may be confused

.github/workflows/js_hooks_test.yml
sublimator: I started working on extracting an `action` in https://github.com/Xahau/xahaud/pull/450

I'll make a PR to this once that's sorted we can just use that

sublimator: @dangell7 wasn't a fan of touching the rippled shared clang-format workflow, in anticipation of merges, but I guess given this PR and #440 both install clang-format, we could use this action: 


```
name: Install Clang-Format Ubuntu
description: Installs a specific version of clang-format on Ubuntu 20.04 and newer
inputs:
  clang-version:
    description: "Version of Clang-Format to install"
    required: true
    default: "10"

runs:
  using: "composite"
  steps:
    - name: Install Clang-Format
      run: |
        codename=$(lsb_release --codename --short)

        # Use modern key management for Ubuntu 20.04+ compatibility
        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /usr/share/keyrings/llvm-archive-keyring.gpg >/dev/null

        # Update LLVM repository sources
        echo "deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-${{ inputs.clang-version }} main" | sudo tee /etc/apt/sources.list.d/llvm.list
        echo "deb-src [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-${{ inputs.clang-version }} main" | sudo tee -a /etc/apt/sources.list.d/llvm.list

        # Update and install Clang-Format
        sudo apt-get update
        sudo apt-get install -y clang-format-${{ inputs.clang-version }}

        # Verify installation
        clang-format-${{ inputs.clang-version }} --version
      shell: bash
```

sublimator: For the record, this is using clang-format 18 which seems to be fine for the file formatted by this.

dangell7:
I'm not sure how valuable the workflow is. We barely touch the test file once its merged. Right now, you will know its different because of the merge conflict. 

I approve the bash changes but I really dont think the CI is necessary to run every single time when it will be touched so little.

tequdev:
> I'm not sure how valuable the workflow is. We barely touch the test file once its merged. Right now, you will know its different because of the merge conflict.
> 
> I approve the bash changes but I really dont think the CI is necessary to run every single time when it will be touched so little.

I will create a separate script-only PR.

tequdev:
- https://github.com/Xahau/xahaud/pull/464

sublimator:
> I really dont think the CI is necessary to run every single time when it will be touched so little

I'll say that yesterday I forgot to update the wasm in one of the PRs, and then it seems one of the tests failed because of that. Of course the beast unit test framework, for some reason just prints that there was 1 failure rather than helpfully listing WHICH test actually failed and at what __file__/__line__. 

This workflow would have helped me more quickly recognize what the issue was. 

I don't think it's strictly necessary either, but it's definitely a small quality of life improvement, and you can set `paths` in the `on` config to make sure it only runs when it changes.

sublimator:
I must note that I've successfully brainwashed myself into referring to the qjsc as "wasm" :)

sublimator:
Closing in favour of #464

PR #455, opened 2025-03-07
Cancel Stale GA Builds 

sublimator:
* Sets the concurrency group to the active branch/PR (ref) and workflow
  * New commits to this group will cancel prior workflow runs
* Adds a cleanup script and always() executes it in the build job
  * release-build.sh writes commands to this to stop the container 
* Leaves the cached builds disabled and TBD but provides required teardown structure
* Checks out to a unique path so mounted folders are unique

```
I remember the issue now. The docker containers are being cached so if you 1 user makes a commit to 2 branches the docker container used the cache from one to test another.

Push Commit Branch 1 -> Triggers Build
Push Commit Branch 2 - Triggers Build (waits)
Branch 1 Build will finish, Branch 2 will start to build.
Branch 2 Build will finish, Brach 1 will test, but tests using Branch 2 Build.
```

This issue MAY be solved by the unique checkout path introduced

### DIFF
--- a/.github/workflows/build-in-docker.yml
+++ b/.github/workflows/build-in-docker.yml
@@ -2,9 +2,9 @@ name: Build using Docker
 
 on:
   push:
-    branches: [ "dev", "candidate", "release", "jshooks" ]
+    branches: ["dev", "candidate", "release", "jshooks"]
   pull_request:
-    branches: [ "dev", "candidate", "release", "jshooks" ]
+    branches: ["dev", "candidate", "release", "jshooks"]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-in-docker.yml
```diff
@@ -7,32 +7,50 @@ on:
...
   checkout:
     runs-on: [self-hosted, vanity]
     steps:
-    - uses: actions/checkout@v4
-      with:
-        clean: false
+      - uses: actions/checkout@v4
+        with:
+          path: ${{ github.sha }}
+          clean: true  # ensure clean checkout
```
sublimator: AI drivel comment 

It's going to be a per commit unique path anyway ...

sublimator: Although, I do wonder about what conditions could cause a clobbering here. I'm not familiar enough with self-hosted runners and how concurrency works. I guess you could push the same commit at the top of different branches.

release-builder.sh
```diff
@@ -19,7 +17,26 @@ if [[ "$GITHUB_REPOSITORY" == "" ]]; then
...
+JOB_CLEANUP_SCRIPT=${JOB_CLEANUP_SCRIPT:-/dev/null}
+NORMALIZED_WORKFLOW=$(echo "$GITHUB_WORKFLOW" | tr -c 'a-zA-Z0-9' '-')
+NORMALIZED_REF=$(echo "$GITHUB_REF" | tr -c 'a-zA-Z0-9' '-')
+CONTAINER_NAME="xahaud_cached_builder_${NORMALIZED_WORKFLOW}-${NORMALIZED_REF}"
+
+# Check if the container is already running
+if docker ps --format '{{.Names}}' | grep -q "^${CONTAINER_NAME}$"; then
+    echo "⚠️ A running container (${CONTAINER_NAME}) was detected."
+
+    if [[ "$EXIT_IF_CONTAINER_RUNNING" -eq 1 ]]; then
```
sublimator, line 32: This was helpful to make sure the cleanup was actually working, left the scaffolding in behind a flag

PR #457, opened 2025-03-10
Outgoing/Incoming HookOn

tequdev:

Instead of making the Outgoing/Incoming txn check within the Hook code, this makes it possible to check it outside the Hook code by defining it in the HookDefinition.

### Context of Change

- Add OutgoingHookOn/IncomingHookOn field

- In the HookDefinition field, one of the following fields is set
    1. HookOn
    2. OutgoingHookOn and IncomingHookOn

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -1222,6 +1262,13 @@ SetHook::setHook()
         std::optional<uint256> newHookOn;
         std::optional<uint256> defHookOn;
 
+        std::optional<uint256> oldOutgoingHookOn;
```
sublimator:
```c++
std::optional<uint256> oldHookOn;
...
std::optional<uint256> oldOutgoingHookOn;
std::optional<uint256> oldIncomingHookOn;
```

These never seem to be actually accessed

sublimator: Would definitely make the code a bit easier to read if they were

tequdev: removed

RichardAH:
I think this is a good idea but should we perhaps just add one new field OutgoingHookOn. If it's absent then HookOn is used for both (as you've already coded here) but if the new field is present then outgoing is filtered through it. A little bit less confusing than adding two new fields?

tequdev:
> I think this is a good idea but should we perhaps just add one new field OutgoingHookOn. If it's absent then HookOn is used for both (as you've already coded here) but if the new field is present then outgoing is filtered through it. A little bit less confusing than adding two new fields?

When considering a scenario where a HookDefinition created by another account has OutgoingHookOn set,
how can we distinguish between:
1. The case where you want to use the same value for OutgoingHookOn as Definition in your own Hook object
2. The case where you want to use only HookOn for a common value for both incoming and outgoing without setting OutgoingHookOn

RichardAH:
> When considering a scenario where a HookDefinition created by another account has OutgoingHookOn set, how can we distinguish between:
> 
> 1. The case where you want to use the same value for OutgoingHookOn as Definition in your own Hook object
> 2. The case where you want to use only HookOn for a common value for both incoming and outgoing without setting OutgoingHookOn

I think if OutgoingHookOn is specified in the definition and you install that hook, but don't specify it, then you'd use the definition's OutgoingHookOn. It's up to you to know what's in the definition. If you want your HookOn field to be the same as OutgoingHookOn then the safest thing to do is actually specify a OutgoingHookOn (identical to HookOn) every time. The backwards compatibility mode is only there to prevent stuff from breaking. People shouldn't use it unless they have to.

tequdev:
@RichardAH 

Do you have any thoughts on using IncomingHookOn instead of OutgoingHookOn?

The Functional Hooks currently under development can only be called from Incoming Txn, so I think it would be better to use the name IncomingHookOn here for this feature.

RichardAH:
> @RichardAH
> 
> Do you have any thoughts on using IncomingHookOn instead of OutgoingHookOn?
> 
> The Functional Hooks currently under development can only be called from Incoming Txn, so I think it would be better to use the name IncomingHookOn here for this feature.

I think it fine as long as its well documented. So in the absence of IncomingHookOn field the HookOn field is both (incoming and outgoing) but if you add IncomingHookOn then HookOn only becomes the outgoing.

RichardAH:
Suggest changing names to HookOnIncoming HookOnOutgoing... following general convention of "name" "action" in both hook apis and transactors

src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -448,12 +454,45 @@ SetHook::validateHookSetEntry(SetHookCtx& ctx, STObject const& hookSetObj)
...
             {
-                JLOG(ctx.j.trace())
-                    << "HookSet(" << hook::log::HOOKON_MISSING << ")["
-                    << HS_ACC()
-                    << "]: Malformed transaction: SetHook must include "
-                       "sfHookOn when creating a new hook.";
-                return false;
+                if (!ctx.rules.enabled(featureHookOnV2) ||
+                    !(hookSetObj.isFieldPresent(sfHookOnOutgoing) &&
+                      hookSetObj.isFieldPresent(sfHookOnIncoming)))
```
RichardAH: Negative A or Negative (B and C) is not intuitive. Prefer positive rather than negative logic, or apply De morgans law (Negative A or Negative B or Negative C)  for ease of reading and to help future programmers not make a logic error here. 

Best practice is make the condition bigger and easier to read like this:
```
if (enabled(featureHookOnV2))
{
  if (isFieldPresent(sfHookOnOutgoing) && isFieldPresent(sfHookOnIncoming))
  {
    // pass, this is valid in v2
  }
  else
  {
      // error here
  }
}
```

PR #468, opened 2025-03-14
Add missing JS_FreeValue calls for early returns

sublimator:
### 🚩 **Issue / Context**
This PR addresses potential memory leaks and incorrect handling of return values in QuickJS-based transaction serialization (`JS_JSONStringify` calls), particularly around the correct freeing of JavaScript values and C-string conversions.

### ✅ **What Has Changed?**

- **Explicit handling of `JS_JSONStringify` return values**:
  - `JS_JSONStringify` can return:
    - A valid string (`JSValue` requiring GC cleanup).
    - An exception (`JS_EXCEPTION`), represented as a special tagged union value (`JS_MKVAL(JS_TAG_EXCEPTION, 0)`), which doesn't require cleanup.
    - JavaScript's `undefined`, represented similarly by a tagged value (`JS_MKVAL(JS_TAG_UNDEFINED, 0)`), also requiring no cleanup.
  - Added proper checks for both `JS_IsException` and `JS_IsUndefined`, handling these special tagged values gracefully and safely.

- **Ensuring correct memory management**:
  - Calls to `JS_FreeCString` and `JS_FreeValue` are now explicitly placed **after every successful conversion** to avoid leaks.
  - Added defensive checks to prevent memory leaks upon early returns (such as large string size limits).

### 🔍 **Explanation of `JS_MKVAL` Values**

`JS_MKVAL` creates lightweight, non-garbage-collected QuickJS "value types":

```c
#define JS_MKVAL(tag, val) (JSValue){ (JSValueUnion){ .int32 = val }, tag }
```

- These are **value types**: they're copied directly on assignment (no pointers involved), meaning they require **no explicit cleanup** (no `JS_FreeValue` needed).
- Examples: `JS_EXCEPTION`, `JS_UNDEFINED`, and integer primitives.
- Confirmed by testing and valgrind analysis, there's no memory leak risk with these special values.

### ⚠️ **Important note about `JS_JSONStringify`**

`JS_JSONStringify` will explicitly return:
- **`JS_EXCEPTION`** when there's an internal error (e.g., cyclic structures or invalid replacer).
- **`JS_UNDEFINED`** when it can't serialize a given input (e.g., `undefined`, function, or symbol).

Both are special `JS_MKVAL` values (value types), hence no GC cleanup required.

### 🚀 **Summary of PR Changes**

- ✅ **Bugfixes:** Memory leak fixes by explicitly cleaning up returned strings and JSValues.
- ✅ **Robustness:** Proper handling of special QuickJS return values (exception, undefined).
- ✅ **Testing:** Confirmed correctness via explicit valgrind checks and careful manual reviews.


src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1739,12 +1739,21 @@ DEFINE_JS_FUNCTION(int64_t, trace, JSValue msg, JSValue data, JSValue as_hex)
...
+            const char* cstr = JS_ToCStringLen(ctx, &len, sdata);
+            if (len > 1023)
+                len = 1023;
+            out += std::string(cstr, len);
+            JS_FreeCString(ctx, cstr);
+            JS_FreeValue(ctx, sdata);
+        }
+        else
+        {
+            // No need to free value types
```
sublimator: Maybe want to handle this error better .... 
`out += "<failed-to-stringify>"` or some such

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -5114,15 +5123,20 @@ DEFINE_JS_FUNCTION(JSValue, emit, JSValue raw_tx)
         // stringify it
         JSValue sdata =
             JS_JSONStringify(ctx, raw_tx, JS_UNDEFINED, JS_UNDEFINED);
-        if (JS_IsException(sdata))
+        if (JS_IsException(sdata) || JS_IsUndefined(sdata))
```
sublimator: Maybe just use `!JS_IsString` ?

src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -1739,12 +1739,21 @@ DEFINE_JS_FUNCTION(int64_t, trace, JSValue msg, JSValue data, JSValue as_hex)
...
+            const char* cstr = JS_ToCStringLen(ctx, &len, sdata);
+            if (len > 1023)
+                len = 1023;
+            out += std::string(cstr, len);
+            JS_FreeCString(ctx, cstr);
+            JS_FreeValue(ctx, sdata);
+        }
+        else
+        {
+            out += "<could not display data>";
```
sublimator, line 1755: matches existing "&lt;could not display hex&gt;"

```c++
 if (JS_IsBool(as_hex) && !!JS_ToBool(ctx, as_hex))
    {
        auto in = FromJSIntArrayOrHexString(ctx, data, 64 * 1024);
        if (in.has_value())
        {
            if (in->size() > 1024)
                in->resize(1024);
            out += strHex(*in);
        }
        else
            out += "<could not display hex>";
    }
```

sublimator:
merged manually

PR #481, opened 2025-03-24
Fix to return zero XFL as bigint

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell7:
I cant remember exactly why I did this but will doing `float_set(...) < 0` work? I think it wont right?

tequdev:
> I cant remember exactly why I did this but will doing `float_set(...) < 0` work? I think it wont right?

Yes, those tests exist and have been successful at least locally.

```cpp
                ASSERT(float_set(void 0, 0) === INVALID_ARGUMENT)
                ASSERT(float_set(0, void 0) === INVALID_ARGUMENT)
                ASSERT(float_set(2147483648, 0) === INVALID_ARGUMENT)
                ASSERT(float_set(-5, 0) === 0n)
                ASSERT(float_set(50, 0) === 0n)
                ASSERT(float_set(-50, 0) === 0n)
                ASSERT(float_set(0, 0) === 0n)
                ASSERT(float_set(-97, 1) === INVALID_FLOAT)
                ASSERT(float_set(97, 1) === INVALID_FLOAT)
```

dangell7:
> > I cant remember exactly why I did this but will doing `float_set(...) < 0` work? I think it wont right?
> 
> Yes, those tests exist and have been successful at least locally.
> 
> ```c++
>                 ASSERT(float_set(void 0, 0) === INVALID_ARGUMENT)
>                 ASSERT(float_set(0, void 0) === INVALID_ARGUMENT)
>                 ASSERT(float_set(2147483648, 0) === INVALID_ARGUMENT)
>                 ASSERT(float_set(-5, 0) === 0n)
>                 ASSERT(float_set(50, 0) === 0n)
>                 ASSERT(float_set(-50, 0) === 0n)
>                 ASSERT(float_set(0, 0) === 0n)
>                 ASSERT(float_set(-97, 1) === INVALID_FLOAT)
>                 ASSERT(float_set(97, 1) === INVALID_FLOAT)
> ```

What about < 0? Those are === 0n. I cant remember exactly. Its fine if there is some other type of handling we need to do but just should be aware and maybe mention if for c hook developers.

tequdev:
> What about < 0? Those are === 0n. I cant remember exactly. Its fine if there is some other type of handling we need to do but just should be aware and maybe mention if for c hook developers.

Do you mean `float_set(-5, 0) < 0`? so `0n < 0`? it's `false`

dangell7:
> > What about < 0? Those are === 0n. I cant remember exactly. Its fine if there is some other type of handling we need to do but just should be aware and maybe mention if for c hook developers.
> 
> Do you mean `float_set(-5, 0) < 0`? so `0n < 0`? it's `false`

I think so

PR #486, opened 2025-04-12
feat(catalogue): add cli commands and fix file_size

sublimator:


src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -72,6 +72,25 @@ static constexpr uint16_t CATALOGUE_COMPRESS_LEVEL_MASK =
 static constexpr uint16_t CATALOGUE_RESERVED_MASK =
     0xF000;  // Bits 12-15: reserved
 
+std::string
+formatFileSize(uint64_t bytes)
+{
+    static const char* units[] = {"B", "KB", "MB", "GB", "TB", "PB"};
```
RichardAH: image
Best practice to use the IEC standard imho... KiB etc...

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -1129,7 +1152,8 @@ doCatalogueLoad(RPC::JsonContext& context)
     jvResult[jss::ledger_count] =
         static_cast<Json::UInt>(header.max_ledger - header.min_ledger + 1);
     jvResult[jss::ledgers_loaded] = static_cast<Json::UInt>(ledgersLoaded);
-    jvResult[jss::file_size] = Json::UInt(file_size);
+    jvResult[jss::file_size_human] = formatFileSize(file_size);
+    jvResult[jss::file_size] = std::to_string(file_size);
```
RichardAH, line 1299: since we're here maybe we should add a projected filesize? this would be the average size of a delta multiplied by the number of deltas plus the size of the first ledger

sublimator:
ACK

PR #497, opened 2025-04-23
ServerDefinitions: Fix SetRemarks & Flags

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/rpc/handlers/ServerDefinitions.cpp
+++ b/src/ripple/rpc/handlers/ServerDefinitions.cpp
@@ -396,6 +396,8 @@ private:
                 return "SetRegularKey";
             if (inp == "HookSet")
                 return "SetHook";
+            if (inp == "RemarksSet")
+                return "SetRemarks";
             return inp;
         };
 


src/ripple/protocol/TxFlags.h
```diff
@@ -192,7 +192,9 @@ enum ClaimRewardFlags : uint32_t {
 constexpr std::uint32_t const tfClaimRewardMask = ~(tfUniversal | tfOptOut);
 
 // Remarks flags:
-constexpr std::uint32_t const tfImmutable = 1;
+enum SetRemarksFlags : uint32_t {
```
tequdev: Since this is not a transaction flag, it is better to leave it unchanged.

src/ripple/rpc/handlers/ServerDefinitions.cpp
```diff
@@ -421,6 +423,7 @@ class Definitions
         addFlagsToJson<NFTokenMintFlags>(ret, "NFTokenMint");
         addFlagsToJson<NFTokenCreateOfferFlags>(ret, "NFTokenCreateOffer");
         addFlagsToJson<ClaimRewardFlags>(ret, "ClaimReward");
+        addFlagsToJson<SetRemarksFlags>(ret, "SetRemarks");
```
tequdev: tfImmutable is a flag used in the sfRemark field, not a transaction flag, so adding it here may be wrong.

dangell7: good point reverted

PR #499, opened 2025-04-25
Suppress build warning introduced in Catalogue

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/test/rpc/Catalogue_test.cpp
+++ b/src/test/rpc/Catalogue_test.cpp
@@ -401,18 +401,8 @@ class Catalogue_test : public beast::unit_test::suite
                 sourceLedger->info().accepted == loadedLedger->info().accepted);
 
             // Check SLE counts
-            std::size_t sourceCount = 0;
-            std::size_t loadedCount = 0;
-
-            for ([[maybe_unused]] auto const& sle : sourceLedger->sles)
-            {
-                sourceCount++;
-            }
-
-            for ([[maybe_unused]] auto const& sle : loadedLedger->sles)
-            {
-                loadedCount++;
-            }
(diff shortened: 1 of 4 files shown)


src/test/rpc/Catalogue_test.cpp
```diff
@@ -404,12 +404,12 @@ class Catalogue_test : public beast::unit_test::suite
             std::size_t sourceCount = 0;
             std::size_t loadedCount = 0;
 
-            for (auto const& sle : sourceLedger->sles)
+            for ([[maybe_unused]] auto const& sle : sourceLedger->sles)
```
sublimator: std::ranges::distance ?

sublimator (dismissed):

src/test/rpc/Catalogue_test.cpp
tequdev: fixed

src/test/rpc/Catalogue_test.cpp
```diff
@@ -829,7 +831,7 @@ class Catalogue_test : public beast::unit_test::suite
         {
             auto result = env.client().invoke(
                 "catalogue_status", Json::objectValue)[jss::result];
-            std::cout << to_string(result) << "\n";
+            // std::cout << to_string(result) << "\n";
```
sublimator, line 834: just remove?

tequdev: I have made it so that you only need to comment it out during testing. I can delete it if you wish.

sublimator: It really doesn't bother me at all

sublimator (approved):

PR #502, opened 2025-04-29
refactor KEYLET_LINE in utils_keylet

tequdev:
Fixes the use of high and low in variable names, as these are determined by ripple::keylet::line processing.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -3145,15 +3145,15 @@ DEFINE_HOOK_FUNCTION(
                 if (a == 0 || b == 0 || c == 0 || d == 0 || e == 0 || f == 0)
                     return INVALID_ARGUMENT;
 
-                uint32_t hi_ptr = a, hi_len = b, lo_ptr = c, lo_len = d,
+                uint32_t acc1_ptr = a, acc1_len = b, acc2_ptr = c, acc2_len = d,
```
sublimator, line 3153: Is this because it doesn't expect the accounts presorted ?

tequdev: Yes, in actually here, it does not matter if a high account is set to lo or a low account is set to hi.

PR #508, opened 2025-05-01
Functional Hooks [DO NOT MERGE]

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

- [x] Public API: New feature (new methods and/or new fields)


RichardAH:
@tequdev are you still of the opinion this is a good idea? Seems to have a fair amount of overlap with much simpler NamedHooks?

tequdev:
> @tequdev are you still of the opinion this is a good idea? Seems to have a fair amount of overlap with much simpler NamedHooks?

To list a few other advantages, there are things like the following, though they might be solvable in ways other than FunctionalHook.

1. Since multiple functions can exist within a single Wasm code, FunctionalHooks can handle complex features with just one hook, whereas NamedHooks would require installing multiple hooks. There are definitely cases where combining them into a single hook is better, as with NamedHooks, you'd also need to verify what happens if only one hook is installed.

2. It uses FunctionParameter, which provides more readable parameter names and values compared to HookParameter. Also, since these values contain type information, you won't need to do as much value validation or conversion within the hook itself.

PR #509, opened 2025-05-06
chore: fix warnings

sublimator:
```
git checkout origin/tmp-conan -- \
    src/test/core/SociDB_test.cpp \
    src/ripple/app/rdb/backend/RWDBDatabase.h \
    src/ripple/overlay/impl/Handshake.cpp
```


dangell7 (dismissed):

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -239,8 +239,7 @@ verifyHandshake(
             throw std::runtime_error("Invalid server domain");
     }
 
-    // Check the network. Omitting Network-ID (on either side ours, or theirs)
-    // means NID=0
+    if (auto const iter = headers.find("Network-ID"); iter != headers.end())
```
dangell7, line 242: This doesn't seem right. Its synced code but it might have been merged correctly. These 2 calls are redundant

dangell7:
Changes in Handshake.cpp need an extra set of eyes. That was a bad merge. This is now exactly whats in ripple(d) but still want to make sure were not breaking anything.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -241,18 +241,12 @@ verifyHandshake(
...
     if (auto const iter = headers.find("Network-ID"); iter != headers.end())
     {
-        uint32_t peer_nid = 0;
-        if (auto const iter = headers.find("Network-ID"); iter != headers.end())
-        {
-            if (!beast::lexicalCastChecked(
-                    peer_nid, std::string(iter->value())))
-                throw std::runtime_error("Invalid peer network identifier");
-        }
+        std::uint32_t nid;
```
sublimator: ooof

RichardAH: you gonna add a {0} here? (defensively) then we can merge?

sublimator: @dangell7 ? 

I think this was something from the Conan branch 
Looks like you took over this PR

RichardAH: I just updated it to current dev branch, do we need this pr or not?

dangell7 (dismissed):

src/ripple/overlay/impl/Handshake.cpp
dangell7: @RichardAH should be ready.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -239,19 +239,14 @@ verifyHandshake(
...
-                throw std::runtime_error("Invalid peer network identifier");
-        }
+        std::uint32_t nid{0};
+
+        if (!beast::lexicalCastChecked(nid, std::string(iter->value())))
+            throw std::runtime_error("Invalid peer network identifier");
 
-        uint32_t our_nid = networkID ? *networkID : 0;
-        if (peer_nid != our_nid)
+        if (networkID && nid != *networkID)
```
RichardAH: this code has a different meaning than what it replaces... should be `if (!networkID && nid != 0 || networkID && nid != *networkID)` shouldn't it?

dangell7: Yes, your right. Updated

If peer sends no Network-ID header but local networkID is set: PR version passes, Suggestion version fails
If peer sends Network-ID but local networkID is unset: PR version passes, Suggestion version fails

sublimator:
We don't even /really/ need this any more, do we?!
It's a relic from a time when the conan branch had stricter warnings enabled, and we wanted to separate some changes in a separate PR iirc

PR #514, opened 2025-05-29
Fix Invalid Tx flags

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell7:
There are other transactions that don't have the right flag rejection. Can you update those as well. There is a ripple PR that addresses this.

tequdev:
> There are other transactions that don't have the right flag rejection. Can you update those as well. There is a ripple PR that addresses this.

Added check for SetSignerList.

Now using the same Amendment name as rippled.
When syncing Creadential, it is better not to use the amendment block of credentials flag

dangell7:
@tequdev Should we merge this into 2.4.0?

tequdev:
> @tequdev Should we merge this into 2.4.0?

There is no need to into 2.4.0.
The Credential flag addressed in rippled is directly merged into featureCredential in sync-2.4.0

PR #516, opened 2025-06-06
Make release build use conan deps where possible and hbb 4.0.1

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


dangell7 (dismissed):

tequdev:
As of the latest build of this PR (2025.7.9-HEAD+1952), I have confirmed that the following three commands work (with `xahaud —version`).

And in WSL(Ubuntu 20.04)

## Debian 10

```bash
docker run --rm --platform linux/amd64 debian:10 bash -c "apt-get update && apt-get install -y wget && wget -O xahaud https://build.xahau.tech/2025.7.9-HEAD%2B1952 && chmod +x xahaud && ./xahaud --version"
```

## Ubuntu 20.04

```bash
docker run --rm --platform linux/amd64 ubuntu:20.04 bash -c "apt-get update && apt-get install -y wget && wget -O xahaud https://build.xahau.tech/2025.7.9-HEAD%2B1952 && chmod +x xahaud && ./xahaud --version"
```

## Rocky Linux 8 (RHEL8 Compatible)

```bash
docker run --rm --platform linux/amd64 rockylinux:8 bash -c "yum install -y wget && wget -O xahaud https://build.xahau.tech/2025.7.9-HEAD%2B1952 && chmod +x xahaud && ./xahaud --version"
```

Builds/CMake/deps/WasmEdge.cmake
```diff
@@ -1,84 +1 @@
...
-#RH NOTE: some compilers / versions of some libraries need these, most don't
-
-find_library(XAR_LIBRARY NAMES xar)
-if(XAR_LIBRARY)
-  target_link_libraries(ripple_libs INTERFACE ${XAR_LIBRARY})
-else()
-  message(WARNING "xar library not found... (only important for mac builds)")
-endif()
-add_library (wasmedge::wasmedge ALIAS wasmedge)
+find_package(wasmedge REQUIRED)
```
sublimator, line 1: Did we check using `ldd` that this is not using the conan shared library?

sublimator: I feel like the libraries made available via the conan toolchain file would probably take precedence here?

tequdev: It has been confirmed that wasmedge is used as a static lib in xahaud binaries created with release-builder.

WasmEdge.cmake is rewritten by build-core.sh. (This is an existing process.)
https://github.com/tequdev/xahaud/blob/0badfd3c9e16de5ef4a4f85143a5f18e0cad1780/build-core.sh#L30-L39

sublimator: Oh, I thought I had seen something about copying over that file, but then when I looked before quickly I couldn't see it

Thanks

release-builder.sh
```diff
@@ -50,25 +49,152 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
+        -DLLVM_LIBRARY_DIR=/usr/lib64/llvm14/lib/ \
+        -DCMAKE_INSTALL_PREFIX=/usr/lib64/llvm14/ \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_EXE_LINKER_FLAGS=\"\$CMAKE_EXE_LINKER_FLAGS\" && \
+    make -j${BUILD_CORES} install && \
+    ln -s /usr/lib64/llvm14/lib/include/lld /usr/include/lld && \
+    cp /usr/lib64/llvm14/lib/liblld*.a /usr/local/lib/ && \
+    cd /tmp && rm -rf lld-* libunwind-*"
+
+# Build and install WasmEdge
```
sublimator: https://github.com/tequdev/xahaud/pull/4

This would have satisfied my curiousity before

release-builder.sh
```diff
@@ -50,25 +49,158 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
 else
   echo "No static container, build on temp container"
   rm -rf release-build;
   mkdir -p release-build;
 
+  docker volume create cache-volume
+
+  # Create inline Dockerfile with environment setup for build-full.sh
+  DOCKERFILE_CONTENT=$(cat <<'DOCKERFILE_EOF'
+FROM ghcr.io/phusion/holy-build-box:4.0.1-amd64
```
sublimator, line 65: You said 4.0.2 is out ?

sublimator: https://github.com/phusion/holy-build-box/releases/tag/rel-4.0.2

tequdev: The latest version is 4.0.1, and it seems that the 4.0.2 has not been released yet.

https://github.com/phusion/holy-build-box/pkgs/container/holy-build-box

release-builder.sh
```diff
@@ -50,25 +49,159 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
   docker start $CONTAINER_NAME
-  docker exec -i $CONTAINER_NAME /hbb_exe/activate-exec bash -x /io/build-core.sh "$GITHUB_REPOSITORY" "$GITHUB_SHA" "$BUILD_CORES" "$GITHUB_RUN_NUMBER"
+  docker exec -i $CONTAINER_NAME /hbb_exe/activate-exec bash -c "source /opt/rh/gcc-toolset-11/enable && bash -x /io/build-core.sh '$GITHUB_REPOSITORY' '$GITHUB_SHA' '$BUILD_CORES' '$GITHUB_RUN_NUMBER'"
   docker stop $CONTAINER_NAME
 else
   echo "No static container, build on temp container"
   rm -rf release-build;
   mkdir -p release-build;
 
+  docker volume create cache-volume
```
sublimator: If this is running on an actions runner with persistent state, which I believe the self-hosted runners qualify as, if there's no volume it will succeed but if it's already existing it will error and the build will stop (`set -ex`) 

It may also pay to have a more descriptive name, "gh-actions-conan-ccache-cache" (or whatever takes your fancy) given the machines may be used for other projects?

build-core.sh
```diff
@@ -38,12 +37,21 @@ target_link_libraries (ripple_libs INTERFACE wasmedge)
...
+
+git config --global --add safe.directory /io &&
 git checkout src/ripple/protocol/impl/BuildInfo.cpp &&
-sed -i s/\"0.0.0\"/\"$(date +%Y).$(date +%-m).$(date +%-d)-$(git rev-parse --abbrev-ref HEAD)+$4\"/g src/ripple/protocol/impl/BuildInfo.cpp &&
+sed -i s/\"0.0.0\"/\"$(date +%Y).$(date +%-m).$(date +%-d)-$(git rev-parse --abbrev-ref HEAD)$(if [ -n "$4" ]; then echo "+$4"; fi)\"/g src/ripple/protocol/impl/BuildInfo.cpp &&
 cd release-build &&
-cmake .. -DCMAKE_BUILD_TYPE=Release -DBoost_NO_BOOST_CMAKE=ON -DLLVM_DIR=/usr/lib64/llvm13/lib/cmake/llvm/ -DLLVM_LIBRARY_DIR=/usr/lib64/llvm13/lib/ -DWasmEdge_LIB=/usr/local/lib64/libwasmedge.a &&
-make -j$3 VERBOSE=1 &&
-strip -s rippled &&
+conan install .. --output-folder . --build missing --settings build_type=$BUILD_TYPE &&
```
sublimator, line 49: We aren't using the soci/snappy from externals which I assume are there for reason, though that brings up questions about cache volume contention. 

See related: #546

release-builder.sh
```diff
@@ -50,25 +49,159 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
+
+  # Create inline Dockerfile with environment setup for build-full.sh
+  DOCKERFILE_CONTENT=$(cat <<'DOCKERFILE_EOF'
+FROM ghcr.io/phusion/holy-build-box:4.0.1-amd64
+
+ARG BUILD_CORES=8
```
sublimator, line 67: It's not really a change here, but "while you're at it", it would be good if this script honoured the BUILD_CORES env, currently if you're not running as a Github Action it's hard coded to 8, which is very wasteful on bigger machines. 

```
BUILD_CORES=$(echo "scale=0 ; `nproc` / 1.337" | bc)

if [[ "$GITHUB_REPOSITORY" == "" ]]; then
  #Default
  BUILD_CORES=8
fi
```

release-builder.sh
tequdev: If the volume already exists, it will be ignored.

https://docs.docker.com/reference/cli/docker/volume/create/#:~:text=If%20you%20specify%20a%20volume%20name%20already%20in%20use%20on%20the%20current%20driver%2C%20Docker%20assumes%20you%20want%20to%20reuse%20the%20existing%20volume%20and%20doesn%27t%20return%20an%20error.

There is no objection to using a different name for the volume name. Considering that it will also be used locally, how about something like `xahau-release-builder-cache`?

build-core.sh
tequdev: I’ll add two `conan export` lines.

release-builder.sh
sublimator: > If the volume already exists, it will be ignored.

Oh :) 

> Considering that it will also be used locally, how about something like xahau-release-builder-cache?

works for me

release-builder.sh
sublimator: https://github.com/tequdev/xahaud/pull/5/files

take it or leave it :) 

I just want ability to set BUILD_CORES plz

tequdev: This setting respects the existing release-builder.sh processing.
How about removing the BUILD_CORE setting processing for cases that are not GitHub Actions?

If you want to build with any number of cores, you can replace the line `BUILD_CORES=$(echo “scale=0 ; `nproc` / 1.337” | bc)`.

sublimator: I'm not really fussed however you do it

I just built on an ubuntu 24.04 machine with many more cores, and didn't realize until a bit of the way through

build-core.sh
sublimator, line 49: reminder: conan has issues with concurrent use of cache

build-core.sh
```diff
@@ -38,12 +37,21 @@ target_link_libraries (ripple_libs INTERFACE wasmedge)
 add_library (wasmedge::wasmedge ALIAS wasmedge)
 message(\"WasmEdge DONE\")
 " > Builds/CMake/deps/WasmEdge.cmake &&
+
+export LDFLAGS="-static-libstdc++"
+export CMAKE_EXE_LINKER_FLAGS="-static-libstdc++"
+export CMAKE_STATIC_LINKER_FLAGS="-static-libstdc++"
```
sublimator: According to Gemini, the Conan toolchain file (conan_toolchain.cmake) takes precedence and defines the build environment for CMake. Since those variables aren't explicitly passed into the CMake
  configuration (via the -D flag), they are effectively ignored for the final executable build.

```
 ldd release-build/xahaud       
        linux-vdso.so.1 (0x00007ffc1c1de000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007bebc6c00000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007bebc703d000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007bebc7038000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007bebc7033000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007bebc6f4a000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007bebc6800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007bebc704b000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007bebc6f1a000)
```

sublimator:
```
cmake .. \
    -G Ninja \
    -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
    -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake \
    -DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++" \
    -DCMAKE_STATIC_LINKER_FLAGS="-static-libstdc++" \
    -DLLVM_DIR=$LLVM_DIR \
    -DWasmEdge_LIB=$WasmEdge_LIB \
    -Dxrpld=TRUE \
    -Dtests=TRUE &&
```

release-builder.sh
sublimator: > even BUILD_CORES=${BUILD_CORES:-8}
 
Bamn, build cores as arg destroys docker caching layer btw :/

build-core.sh
sublimator:
```
[62/673] Building CXX object CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/TER.cpp.o
[63/673] Linking C static library src/ed25519-donna/libed25519.a
FAILED: src/ed25519-donna/libed25519.a 
: && /usr/bin/cmake -E rm -f src/ed25519-donna/libed25519.a && /usr/bin/ar qc src/ed25519-donna/libed25519.a -static-libstdc++ src/ed25519-donna/CMakeFiles/ed25519.dir/ed25519.c.o && /usr/bin/ranlib src/ed25519-donna/libed25519.a && :
/usr/bin/ar: two different operation options specified
[64/673] Building CXX object CMakeFiles/xrpl_core.dir/src/ripple/crypto/impl/RFC1751.cpp.o
[65/673] Building CXX object CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/PublicKey.cpp.o
```

Oh, wait

```
-DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++" \
-DCMAKE_STATIC_LINKER_FLAGS="-static-libstdc++" \
```

***

### CMAKE_EXE_LINKER_FLAGS

* **Tool:** The **linker** (like `ld`, which is usually called by the compiler driver `g++` or `clang++`).
* **Purpose:** To link object files and libraries together to create a final executable or a shared library (`.so`).
* **Correct Usage:** This is the correct place to put linker-specific flags like `-static-libstdc++`.

***

### CMAKE_STATIC_LINKER_FLAGS

* **Tool:** The **archiver** (`ar`).
* **Purpose:** To bundle object files (`.o`) into a static library archive (`.a`). It doesn't perform linking; it just collects files into a single package.
* **Correct Usage:** This variable should **not** contain linker flags. Passing `-static-libstdc++` here is what caused your error because `ar` doesn't understand that flag.

In short, you've diagnosed the problem correctly: a flag meant for the linker was incorrectly passed to the archiver. Moving `-static-libstdc++` from `CMAKE_STATIC_LINKER_FLAGS` to `CMAKE_EXE_LINKER_FLAGS` should resolve the build failure.

sublimator: set_target_properties(rippled PROPERTIES LINK_FLAGS "-static-libstdc++")

Though, it seems we might actually need this in RippledCore.cmake, because `CMAKE_EXE_LINKER_FLAGS` is not being honoured

sublimator: Nor is that, actually

sublimator: Suspecting conan binaries linked against dynamic libstdc++.so.6

sublimator: Seems to be snappy 

```
> ╭─niq@niq-xahau ~/xahau/xahaud-worktrees/xahaud/release-build/build/generators ‹conan-release-builder●› 
  ╰─$ grep -r -- 'stdc++'   
  Snappy-release-x86_64-data.cmake:set(snappy_SYSTEM_LIBS_RELEASE m stdc++)
  Snappy-release-x86_64-data.cmake:set(snappy_Snappy_snappy_SYSTEM_LIBS_RELEASE m stdc++)
  ╭─niq@niq-xahau ~/xahau/xahaud-worktrees/xahaud/release-build/build/generators ‹conan-release-builder●›
```

sublimator: https://gist.github.com/sublimator/23cdb66ca497139fed256e559d9e75e4

sublimator: Let me correct myself and stick only to the hard evidence we have gathered.

  Here are the facts, without any assumptions:

  1. Fact: The sed command that removes stdc++ from Snappy-release-x86_64-data.cmake works. It produces a correctly linked binary. This is proven by your final ldd output.
  2. Fact: The grep command proves that Snappy-release-x86_64-data.cmake is the only file in the entire build directory that contains the incorrect stdc++ system library directive.
  3. Fact: The snappy/conanfile.py is the only recipe we have inspected that explicitly adds a library ("m") to the system_libs property.
  4. Fact: All attempts to fix this by changing global settings (like the Conan profile or command-line flags) have failed.

  Based on these facts, the only logical conclusion is that there is a unique, specific interaction between the snappy recipe and the Conan generator that causes the incorrect file to
   be created.

  You are right to say that my explanation of why the generator does this is an assumption. But we don't need to know the exact "why" to be certain of the "what" and the "where."

  The evidence proves the problem is isolated to the generation of that one file, and therefore, your sed patch is the correct, targeted, and evidence-based solution.

build-core.sh
```diff
@@ -38,12 +37,23 @@ target_link_libraries (ripple_libs INTERFACE wasmedge)
...
+
+git config --global --add safe.directory /io &&
 git checkout src/ripple/protocol/impl/BuildInfo.cpp &&
-sed -i s/\"0.0.0\"/\"$(date +%Y).$(date +%-m).$(date +%-d)-$(git rev-parse --abbrev-ref HEAD)+$4\"/g src/ripple/protocol/impl/BuildInfo.cpp &&
+sed -i s/\"0.0.0\"/\"$(date +%Y).$(date +%-m).$(date +%-d)-$(git rev-parse --abbrev-ref HEAD)$(if [ -n "$4" ]; then echo "+$4"; fi)\"/g src/ripple/protocol/impl/BuildInfo.cpp &&
+conan export external/snappy snappy/1.1.9@ &&
```
sublimator: See #546 
Our conanfile is using 1.1.10 
And the fact that we mispecified this again, is a really good argument for namespacing! 

```
conan export external/snappy snappy/1.1.10@xahaud/stable
conan export external/soci soci/4.0.3@xahaud/stable
```

I don't know about conan 2, but with conan 1, when we are exporting with only "soci/4.0.3@" and our recipe is different to rippled, which also exports using that name, you end up having to export again every time you switch between xahaud and rippled.

Also, as you can see, it's very easy for the builds to silently use the conan version.

build-core.sh
sublimator: I actually had to use "conan remove --locks" at one point when experimenting after canceling a build midway through

build-core.sh
sublimator: I removed completely the `system_libs.append("m")` from the snappy recipe, because `m` was always going to get dragged into the mess, in a system as complex as `{xahau,ripple}d` and it confirms the theory of the `m` triggering some accidental inclusion of `lstdc++`. 

```
> ╭─niq@niq-xahau ~/xahau/xahaud-worktrees/xahaud ‹conan-release-builder●› 
  ╰─$ ldd release-build/xahaud                                   
          linux-vdso.so.1 (0x00007ffd7119f000)
          libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x0000719157dda000)
          librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x0000719157dd5000)
          libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x0000719157dd0000)
          libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000719157ce7000)
          libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000719157a00000)
          /lib64/ld-linux-x86-64.so.2 (0x0000719157de8000)
```

We could just comment it out in our external, add comments as to why, and if we were feeling pedantic enough, explicitly add `m` to our executable targets.

build-core.sh
tequdev: I'll fix this to 1.1.10.

and I'll fix the namespace once the PR is merged.

build-core.sh
```diff
@@ -55,6 +55,7 @@ ninja -j $3 &&
 ccache -s &&
 strip -s rippled &&	
 mv rippled xahaud &&
+libcheck xahaud &&
```
sublimator, line 63: For anyone else curious:
https://github.com/phusion/holy-build-box/blob/master/image/libcheck

build-core.sh
```diff
@@ -38,13 +37,30 @@ target_link_libraries (ripple_libs INTERFACE wasmedge)
 add_library (wasmedge::wasmedge ALIAS wasmedge)
 message(\"WasmEdge DONE\")
 " > Builds/CMake/deps/WasmEdge.cmake &&
+
+export LDFLAGS="-static-libstdc++"
```
sublimator, line 41: It seems to be working, but we should probably be "extending" LDFLAGS:
https://github.com/phusion/holy-build-box/blob/master/image/activate_func.sh#L14-L15

```
export LDFLAGS="$LDFLAGS -static-libstdc++"
```

sublimator: activate_holy_build_box_deps_installation_environment

Though that may not be relevant at all

sublimator: https://github.com/phusion/holy-build-box/blob/master/image/activate_func.sh#L24-L45

Ok, no, `activate_holy_build_box` is setting LDFLAGS too

sublimator: also, comment that is relevant to the conan deps (my assumption) probably good

build-core.sh
sublimator: For record:
https://github.com/tequdev/xahaud/pull/6/files

RichardAH:
Awesome job guys

PR #528, opened 2025-06-27
Conan Release Builder (2.4.0 sync)

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


release-builder.sh
```diff
@@ -1,9 +1,8 @@
-#!/bin/bash -u
+#!/bin/bash
```
sublimator, line 1: Unsure why removed the -u and set -ex yet as haven't read down, but the comment remains

release-builder.sh
```diff
@@ -50,25 +49,187 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
+    make -j${BUILD_CORES} install" && \
+    cp -r include/api/wasmedge /usr/include/ && \
+    cd /tmp && rm -rf WasmEdge* 0.11.2.zip
+
+# Set environment variables
+ENV PATH=/usr/local/bin:$PATH
+
+# Configure ccache and Conan
+RUN /hbb_exe/activate-exec bash -c "ccache -M 10G && \
+    ccache -o cache_dir=/cache/ccache && \
```
sublimator, line 179: Note that unless you configure ccache to use content tracking, afaik it's not very useful in a CI situation due to git checkouts updating the modification times of files


> By default, ccache uses mtimes (modification times) to detect changes. Git checkouts reset timestamps, making ccache think everything changed even if content is identical. Setting compiler_check = content makes ccache hash the actual file contents instead.

release-builder.sh
tequdev: I deleted it because an error occurred when executing locally because some GitHub ENV did not exist in local

release-builder.sh
tequdev: I tried it locally, but I couldn't see any difference in behavior between the default and compiler_check=content.

release-builder.sh
```diff
@@ -50,25 +49,187 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
+    make -j\${BUILD_CORES} && \
+    make install && \
+    ln -sf /usr/local/bin/ar /usr/bin/ar && \
+    cd /tmp && \
+    rm -rf binutils binutils.tar.gz"
+
+# Build and install WasmEdge
+RUN cd /tmp && \
+    ( wget -nc -q https://github.com/WasmEdge/WasmEdge/archive/refs/tags/0.11.2.zip; unzip -o 0.11.2.zip; ) && \
+    cd WasmEdge-0.11.2 && \
```
sublimator, line 151: When you install this and boost like so will conan install --missing will skip these?

sublimator: Also curious why using these over the conan versions ? Specific build flags?

sublimator: '-D_GLIBCXX_USE_CXX11_ABI=0'

But then if Boost/WasmEdge need old ABI, don't ALL the C++ dependencies need it too?

release-builder.sh
```diff
@@ -50,25 +49,187 @@ STATIC_CONTAINER=$(docker ps -a | grep $CONTAINER_NAME |wc -l)
...
+    ./b2 link=static -j${BUILD_CORES} && \
+    ./b2 install && \
+    cd /tmp && \
+    rm -rf boost boost.tar.gz"
+
+ENV BOOST_ROOT=/usr/local/src/boost_1_86_0
+ENV Boost_LIBRARY_DIRS=/usr/local/lib
+ENV BOOST_INCLUDEDIR=/usr/local/src/boost_1_86_0
+
+ENV CMAKE_CXX_FLAGS='-D_GLIBCXX_USE_CXX11_ABI=1'
```
sublimator: New ABI here though?
cmake will honor "CMAKE_CXX_FLAGS" as if its CXX_FLAGS and pass along?

release-builder.sh
tequdev: conan install will not skip those.
However, wasmedge will use what was built here in release-build. boost should be used the one from conan.

To use it as a static lib, building wasmedge here.
The one provided by conan is a shared lib.

sublimator: > The one provided by conan is a shared lib.

I wonder if that can be configured in the conan file somehow

release-builder.sh
sublimator: I mean, it's a moot point when there's nothing cached
It's only when you're able to reuse the container somehow

release-builder.sh
sublimator: > boost should be used the one from conan.

Ideally there's only one version of boost linked, but I confess I'm not an expert on linking so I don't know exactly what could go wrong there

sublimator: Given it's static, it's likely not a problem, is it? Because the final linking isn't done until an actual binary is made (rippled), where it's going to use the conan boost. Maybe some template specialization issues.

tequdev: https://github.com/Xahau/xahaud/pull/516/commits/01a3c4c45160fbbb7d5ebca954d4a0db08a2a71c

release-builder.sh
tequdev: https://github.com/Xahau/xahaud/pull/516/commits/ae1c906e0348664751b852ed1242b5468f7cf089

release-builder.sh
tequdev: removed

PR #533, opened 2025-07-02
Add new keylets to `util_keylet`

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


RichardAH:
I think this needs an amendment because the hook api behaviour is modified and could cause a fork condition?

tequdev:
> I think this needs an amendment because the hook api behaviour is modified and could cause a fork condition?

As for me, I believe this amendment for the fix is unnecessary. In the code before the change, an `INVALID_ARGUMENT` was returned due to the condition `LAST_KLTYPE_V1`, and it never reached `NO_SUCH_KEYLET`.

However, I wanted these to be checked, so I created them as a separate PR.

RichardAH:
> As for me, I believe this amendment for the fix is unnecessary. In the code before the change, an `INVALID_ARGUMENT` was returned due to the condition `LAST_KLTYPE_V1`, and it never reached `NO_SUCH_KEYLET`.
> 
> However, I wanted these to be checked, so I created them as a separate PR.

If I have a hook that calls this api with one of the new keylet types then nodes that have updated will evaluate and return the keylet and nodes that haven't will return INVALID_ARGUMENT? Am I missing something?

tequdev:
> If I have a hook that calls this api with one of the new keylet types then nodes that have updated will evaluate and return the keylet and nodes that haven't will return INVALID_ARGUMENT? Am I missing something?

All new keylets are Amendment Guarded.

And since the refactoring other than the addition of the new Keylet is merged in the following PR, I need to update this PR.

https://github.com/Xahau/xahaud/pull/612/files

PR #534, opened 2025-07-07
Feature Clawback

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -1,0 +1,140 @@
+//------------------------------------------------------------------------------
+/*
+  This file is part of rippled: https://github.com/ripple/rippled
+  Copyright (c) 2023 Ripple Labs Inc.
+
+  Permission to use, copy, modify, and/or distribute this software for any
+  purpose  with  or without fee is hereby granted, provided that the above
+  copyright notice and this permission notice appear in all copies.
+
+  THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+  WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+  MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+  ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+  WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+  ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/tx/impl/Clawback.h>
+#include <ripple/ledger/View.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Indexes.h>
+#include <ripple/protocol/Protocol.h>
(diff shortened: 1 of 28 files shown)


src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // Get the spendable balance. Must use `accountHolds`.
+    STAmount const spendableAmount = accountHolds(
+        view(),
+        holder,
+        clawAmount.getCurrency(),
+        clawAmount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
+
+    return rippleCredit(
```
RichardAH, line 131: if the line is under escrow (i.e. has a locked balance then preferentially try to claw back whatever isn't locked then claw back under the lock (decrement the lock).

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -847,4 +848,62 @@ NFTokenCountTracking::finalize(
...
+    {
+        if (trustlinesChanged != 0)
+        {
+            JLOG(j.fatal()) << "Invariant failed: some trustlines were changed "
+                               "despite failure of the transaction.";
+            return false;
+        }
+    }
+
+    return true;
```
RichardAH, line 1203: add a check for lockedbalance sanity in the invariant checker too

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,1049 @@
...
+
+        // bob should have 800 USD left
+        BEAST_EXPECT(env.balance(bob, USD) == USD(800));
+
+        // gw claws back 800 USD from bob
+        env(claw(gw, bob["USD"](800)));
+        env.close();
+
+        // bob should have 10 USD left.
+        // NOTE: You cannot claw back the locked tokens.
```
RichardAH: I'm pretty sure this isn't enforced in code at the moment? but for regulatory reasons the issuer's clawback right trumps the escrow, so they should be able to clawback under an escrow. decrement the locked balance as needed.

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,1049 @@
...
+        env.require(balance(alice, bob["USD"](-50)));
+
+        // Verify that the account sequence numbers did not advance.
+        BEAST_EXPECT(env.seq(alice) == aliceSeq);
+    }
+
+    void
+    testWithFeats(FeatureBitset features)
+    {
+        // testAllowTrustLineClawbackFlag(features);
```
RichardAH: these are all commented out?

src/ripple/protocol/Feature.h
```diff
@@ -369,6 +369,7 @@ extern uint256 const fixXahauV3;
 extern uint256 const fix20250131;
 extern uint256 const featureHookCanEmit;
 extern uint256 const fixRewardClaimFlags;
+extern uint256 const featureClawback;
```
sublimator: Logic error: More features defined than allocated. Adjust numFeatures in Feature.h.

RichardAH: yup my fault fix please

src/ripple/protocol/TxFormats.h
```diff
@@ -139,6 +139,9 @@ enum TxType : std::uint16_t
     /** This transaction accepts an existing offer to buy or sell an existing  NFT. */
     ttNFTOKEN_ACCEPT_OFFER = 29,
 
+    /** This transaction claws back issued tokens. */
+    ttCLAWBACK = 44,
```
sublimator: This seems to conflict with rippled: 

- Xahaud: ttCLAWBACK = 44
- Rippled: ttXCHAIN_ACCOUNT_CREATE_COMMIT = 44

30, which [rippled is using](https://github.com/XRPLF/rippled/blob/develop/include/xrpl/protocol/detail/transactions.macro#L233) seems to be available

It might make future syncing easier if use the same transaction type id? 
It's using the same amendment name

PR #542, opened 2025-07-10
Fix RWDB Online Delete

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/rdb/backend/RWDBDatabase.h
```diff
@@ -65,9 +65,17 @@ class RWDBDatabase : public SQLiteDatabase
             return {};
 
         std::shared_lock<std::shared_mutex> lock(mutex_);
+        LedgerIndex minSeq = std::numeric_limits<LedgerIndex>::max();
         if (transactionMap_.empty())
             return std::nullopt;
-        return transactionMap_.begin()->second.second->getLgrSeq();
+        for (const auto& [_, txMeta] : transactionMap_)
```
sublimator, line 71: The assumption was the map was sorted by ledger sequence somehow, but it's sorted by random hash, so we need to walk over the entire map.

It seems this function is only really used by SHAMapStoreImp::clearPrior (and the tests) so the O(n) is probably ok

sublimator: Although, it may be better to use the ledgers_ map:

```
struct LedgerData
{
    LedgerInfo info;
    std::map<uint256, AccountTx> transactions;
};

std::map<LedgerIndex, LedgerData> ledgers_;
```

And search for first LedgerData with non-empty transactions map?

PR #546, opened 2025-07-16
fix(ci): export correct snappy version

sublimator:
Snappy export version in CI was not matching that in [conanfile](https://github.com/Xahau/xahaud/blob/dev/conanfile.py#L35)

https://github.com/Xahau/xahaud/blob/849d447a20d920495afac7a3311b75398b5ffed6/conanfile.py#L26-L40

# Summary

This PR fixes the Conan dependency exports to use properly namespaced package references for custom recipes, ensuring consistent dependency resolution across the build system and preventing linking conflicts.

## Changes

- Updated snappy export from `snappy/1.1.9@` to `snappy/1.1.10@xahaud/stable` in GitHub Actions
- Updated soci export to use `soci/4.0.3@xahaud/stable` namespace in both GitHub Actions and build documentation
- Updated `conanfile.py` to reference the namespaced versions of snappy and soci
- Fixed SOCI's boost dependency to use `boost::headers` instead of `boost::boost` to prevent linking all boost libraries and avoid multiple definition errors (e.g., `__cxa_allocate_exception`)

## Impact

These changes:

1. **Prevent linking conflicts**: Using `boost::headers` instead of `boost::boost` avoids pulling in all boost libraries, which was causing multiple definition errors during static linking
2. **Enable parallel development**: The `@xahaud/stable` namespace prevents conflicts when working on both xahaud and rippled projects simultaneously, as they use different SOCI 4.0.3 recipes
3. **Ensure reproducible builds**: Properly namespaced custom recipes guarantee consistent dependency resolution across different environments. This fixes a version mismatch where the build was incorrectly using the Conan Center version of snappy (1.1.9) instead of the exported version (1.1.10)

## Technical Details

The previous configuration was causing linker errors like:

```
/usr/bin/ld.gold: error: /usr/lib/gcc/x86_64-linux-gnu/11/libstdc++.a(eh_alloc.o): multiple definition of '__cxa_allocate_exception'
/usr/bin/ld.gold: /home/runner/.conan/data/boost/1.86.0/_/_/package/.../lib/libboost_stacktrace_from_exception.a(from_exception.o): previous definition here
```

This fix ensures only the necessary boost headers are included rather than linking against all boost libraries.

See also: #529


tequdev (dismissed):
Can you fix this one too?

https://github.com/Xahau/xahaud/blob/849d447a20d920495afac7a3311b75398b5ffed6/BUILD.md#L120

Issue #547, opened 2025-07-17
Catalogue "make quick issue" for Richard re: memory usage


WietseWind:
<img width="677" height="1139" alt="Image" src="https://github.com/user-attachments/assets/6c8905c1-2a16-40e3-bdde-9c96bd0522bf" />

PR #548, opened 2025-07-18
perf(ledger): optimize catalogue loading memory usage and performance

sublimator:
This PR addresses #547 
  
# Catalogue Loading and Persistent Pinned History

## TL;DR

**Problem**: Catalogue loading consumed 200GB RAM for a 53GB file, never freed memory. The old pinned-ledger approach also tried to copy ALL pinned ledgers on EVERY rotation, which does not scale.

**Solution**: `DatabasePinnedImp` provides dual backends: a rotating backend for hot ledgers and a persistent backend for pinned nodes. Type-based routing via `NodeObjectType` keeps pinned history out of rotation entirely, with bounded-memory catalogue loading and clearer operational behavior.

**Journey**: Attempted extensive parallelization (concurrent jobs, pointer-diff flushing, canonicalize-then-flush, JobQueueAdapter). All worked but provided marginal gains. SHAMap COW limitations + SQLite write bottleneck = inescapable. "The Great Simplification" (commit 93a16ac3b): deleted 585 lines, returned to synchronous single-threaded approach. Optimize for simplicity and reviewability.

**What Survived**:
- DatabasePinnedImp architecture
- Cache bypass for pinned ledgers
- Pinned-range-aware publishing and online_delete behavior
- Synchronous standalone execution
- Simple, reviewable code and battle-tested tuning parameters

## DatabasePinned UserGuide

See [here](https://gist.github.com/sublimator/9f55af9be366b7302700f186d3af53ef)

## Initial Report

The catalogue loading feature was working but exhibited severe memory consumption problems:

**Problem Description:**
- Loading a 53GB catalogue file (with max compression) consumed nearly 200GB of RAM
- Memory was never freed between successive file loads
- Pattern observed: "grow grow grow, finish, next one, grow grow grow, etc."
- On a system with 512GB RAM, memory usage would continuously increase

**Memory Statistics During Load:**
```
MemTotal:     528352864 kB (~504 GB)
MemFree:        3502420 kB (~3.3 GB)
MemAvailable: 326320868 kB (~311 GB)
```

**Initial Hypothesis:**
Richard suspected that pinned ledgers might be staying in memory instead of being flushed to the backend storage, despite the design intent that they should be periodically written to disk.

This memory leak behavior made catalogue loading impractical for production use, as it required massive amounts of RAM and the memory was never reclaimed.

## Debugging the Memory Bloat

Initial investigation revealed the memory issue wasn't a simple leak - ledgers were being held in memory by multiple cache layers and pending save operations.

**The dual approach**: Optimize SQLite ledger save jobs AND prevent pinned ledgers from entering caches.

### SQLite Optimization Attempts

Tried batching operations to reduce database calls, combining DELETEs, batching INSERTs, and experimenting with worker limits and synchronous save options.

**Why it was dropped**: Unclear write lock / WAL tuning interactions and inconsistent performance that degraded after initial good peformance. Measurements were difficult - ledger sizes vary and SQLite write performance is unpredictable, especially in a live p2p environment. Safer to keep "battle-tested" code.

### Cache Bypass Strategy

**The discovery**: Pinned ledgers were filling multiple caches throughout the system:
- **AcceptedLedgerCache** - Used by NetworkOPs for publishing
- **LedgerHistory (mLedgerHistory)** - Must contain ledgers for publishing to work
- **NodeStore cache** - A vestigial 2014 workaround that Ripple is removing

### The Rotation Problem Discovery

The existing pinned ledger implementation had a fundamental problem that made it completely impractical:

**The "Wait, What?" Moment**:
- The code was trying to copy ALL pinned ledgers on EVERY database rotation
- Writing FULL trees for each ledger, every single time
- With millions of pinned ledgers, you could be due for the next rotation before the copy even finished
- Essentially: never going to work

**Evolution of the Solution**:

1. **First reaction**: The old `ensurePinnedLedgersInWritable()` approach was clearly hopeless. Parts of that path lingered as dead/zombie code for a while, but the design was ultimately removed rather than repaired.
2. **Maybe make it cheap?**: Try LVM snapshots with append-only NUDB? Prepare database templates via external script?
3. **Wait... why?**: After exploring snapshots and layering approaches, the breakthrough question: **Why tie pinned ledgers to rotation at all?**

**The Solution**:
Just use a completely separate backend that never rotates. DatabasePinnedImp with dual backends - one rotating for hot ledgers, one persistent for pinned nodes. No copying, no complexity, clean separation of concerns.

## The Inescapable Bottleneck: SQLite Writes

After implementing DatabasePinnedImp and solving the rotation problem, extensive optimization attempts were made to speed up catalogue loading through parallelization:

**Optimization Attempts:**
1. **Parallel save jobs** - Queue highly concurrent save operations
2. **Pointer-diff flushing** - Use pointer comparison to identify changed nodes
3. **Canonicalize-then-flush** - Separate CPU work (hashing) in main thread from I/O work (flushing) in background
4. **Async SQL operations** - Queue SQL saves separately from node flushing

**The Realization:**
All these clever optimizations provided marginal improvements at best, because **SQLite write performance is the bottleneck**:

- You can load/build ledgers as fast as you want
- You can flush SHAMap nodes in parallel
- You can optimize COW and memory management
- **But SQLite can't write fast enough to keep up**

**The Problem:**
- Loading ledgers faster just creates a **long tail of pending SQL save jobs**
- These jobs keep ledgers in memory (can't free until saved)
- Memory bloats again despite all the cache bypassing
- SQLite sustained write performance degrades after 20-30 minutes regardless of tuning

**The Live Environment Constraint:**
In a non-standalone (live) environment, any optimization that speeds up catalogue loading is **stealing resources from P2P and transaction processing**:
- Database locks affect consensus operations
- Memory pressure impacts normal node operations
- JobQueue contention slows down validation

**The Pragmatic Conclusion:**
Catalogue loading is a **last resort tool**. Once custom offline tools exist, it likely won't even be used. But if it's there, it should **play nice** with the rest of the system:

- Single-threaded, synchronous approach in standalone mode
- No complex parallelization that adds risk
- Keep battle-tested code/params that's easy to review
- Accept that it's slow - it's meant for rare use

**Final simplification principle:** Remove all the complexity. The bottleneck can't be eliminated, so optimize for code simplicity and reliability instead of marginal speed gains.

## DatabasePinnedImp Architecture

The final solution uses a dual-backend architecture implemented in `DatabasePinnedImp.h/cpp` that routes nodes based on their `NodeObjectType`.

### Design

**Composition Pattern**: `DatabasePinnedImp` inherits from `DatabaseRotating` and uses composition with `DatabaseRotatingImp` to handle hot nodes, while adding a persistent backend for pinned nodes.

**Routing Logic** (`store()` method):
- `pinnedACCOUNT_NODE`, `pinnedTRANSACTION_NODE`, `pinnedLEDGER` → persistent backend (never rotates)
- All other types → rotating backend (rotates via online_delete)

**Fetching**: Tries rotating backends first (hot data), falls back to persistent backend (pinned data). Pinned data is never copied to rotating storage, even with `duplicate=true`.

**Rotation**: Simply delegates to `DatabaseRotatingImp` - only the rotating backend participates in rotation.

### Configuration

Located in `[node_db]` section of config file:

```ini
[node_db]
type=rwdb                    # Rotating backend type (rwdb, NuDB, RocksDB, etc.)
pinned_type=NuDB             # Persistent pinned backend; non-standalone requires NuDB or RocksDB
pinned_path=/path/to/pinned  # Required for filesystem-backed pinned backends
online_delete=2000           # Required - enables rotation to prevent memory OOM
```

**Validation** (in `Config.cpp`):
- `catalogue_load` now rejects unless `[node_db] pinned_type` is configured
- Requires `type` is specified when using `pinned_type`
- Requires `online_delete > 0` (rotation necessary for memory management)
- Requires `pinned_path` for filesystem-backed pinned backends
- In non-standalone mode, `pinned_type` must be a durable backend (`NuDB` or `RocksDB`)

### Key Features

**Clean Separation**: Pinned ledgers bypass rotation entirely - no copying, no complexity from the old `ensurePinnedLedgersInWritable()` approach.

**Type-Based Routing**: Simple switch on `NodeObjectType` in the `store()` method handles everything. No configuration for ranges or explicit pinning decisions at runtime.

**Delegation**: All rotation logic handled by existing `DatabaseRotatingImp` - no duplication of rotation code.

**File Descriptors**: Aggregates `fdRequired()` from both backends for resource management.

### Operational Hardening

- `catalogue_load` now fails early if `pinned_type` is not configured, since loading into rotating storage would be lost on the next rotation.
- Each loaded ledger is pinned before save, and an RAII guard calls `unpinLedger()` if the save path fails before completion.
- The `pendSaveValidated` wait path now treats a broken promise as a hard failure instead of silently continuing.
- `catalogue_load` now validates `min_ledger <= max_ledger` before processing.
- Restored pinned ranges are merged into `complete_ledgers` in a one-shot, restart-safe way via `mPinnedMergedToComplete`.
- Startup is fail-fast when `state.db` says pinned history exists but the currently configured pinned store cannot satisfy the persisted range boundaries.
- Pinned node types are reduced back to serializable hot types before encoding/storage via `isPinnedType()` / `toHotType()`, avoiding type-truncation issues.
- Online-delete range pruning keeps the range-based approach and forward progress while respecting pinned intervals.
- `DatabasePinnedImp::importDatabase()` is explicitly unsupported and now throws a more useful message directing operators to `catalogue_load` or a non-pinned node store.

## SHAMap COW Limitations

During the parallelization experiments, extensive testing revealed fundamental limitations in Ripple's SHAMap Copy-on-Write (COW) implementation that prevented efficient parallel processing. These findings are documented in the now-removed `SHAMapCOW_test.cpp` (1697 lines).

### The Fundamental Theorem

**For efficient tree hashing with cached hashes:**
1. You MUST hash bottom-up (leaves before parents)
2. You MUST cache hashes in nodes (or rehash everything)
3. Caching hashes MODIFIES nodes
4. Shared nodes can't be modified concurrently
5. **Therefore: Hashing shared trees MUST be serial**

### Critical Finding: getHash() Kills Flushing

**The Problem:**
- `getHash()` computes hashes and marks all nodes as clean (cowid=0)
- Once nodes are marked clean, `flushDirty()` returns 0 nodes - nothing to flush
- This happens to ALL maps sharing those nodes (parent AND children)

**Test Results:**
- Before `getHash()`: Can flush 133 nodes from a 3-ledger chain
- After `getHash()`: Can flush 0 nodes - everything marked clean forever

### COW Binary Ownership Problem

**Design:**
- COW uses binary ownership: cowid=0 (clean/shared) or non-zero (dirty/exclusive)
- Can't track deltas between multiple versions
- Snapshots cause all nodes to reset to cowid=0

**Impact on Ledger Chains:**
- Can't use pointer-diff flushing without hashing first
- Can't hash without marking nodes clean
- Can't flush after marking clean
- **Serial dependency:** build → flush → hash, repeat

### Race Conditions

**Threading Issues:**
- Multiple snapshots share most nodes throughout the tree
- All threads hashing snapshots race to cache hashes in the same shared nodes
- Building next ledger while flushing previous ledger = data race
- Even with canonicalization, timing windows cause crashes

**Real Catalogue Loading Scenario:**
- Main thread: Build ledger N+1 from ledger N
- Background thread: Flush ledger N-1
- Both touching shared COW tree structure
- Result: Segfault at ~58% completion

### The "Safe" Pattern (That Worked, But Wasn't Worth It)

**Canonicalize-then-flush approach:**
1. Main thread: Build ledger N
2. Main thread: Canonicalize (walkSubTree with doWrite=false) - computes hashes, marks cowid=0
3. Main thread: Create ledger N+1 snapshot
4. Background thread: Pointer-diff flush of ledger N
5. Main thread: Continue building N+1

**Why it was safe:**
- Canonical nodes have cowid=0 (immutable)
- Pointer comparison is stable
- Background reads while main builds next
- No race conditions

**Why it was abandoned:**
- It DID work, but only provided marginal speed improvement
- The SQLite bottleneck is inescapable - you end up with a long tail of pending SQL jobs
- Loading faster just queues more work, doesn't finish faster overall
- Massive complexity (pointer-diff logic, canonicalization, JobQueue coordination) for minimal gain
- Simplicity and reviewability more valuable than marginal performance that's only thrown away

### Why Partitioning Doesn't Work

**Theory:** Partition tree, hash subtrees in parallel, combine at top

**Reality:**
- Partitioning ONE tree on its first hash could work - subtrees are independent
- But catalogue loading builds CHAINS of ledgers via COW snapshots
- Subsequent ledgers share partitions with previous ledgers
- Those shared partitions can't be hashed concurrently
- You'd only get parallelism on the first ledger, then serial after that
- Defeats the purpose for catalogue loading millions of ledgers

### The Memory Problem
Catalogue loading consumed 200GB RAM for a 53GB file - memory that was never freed. Pinned ledgers were filling AcceptedLedgerCache, LedgerHistory, and NodeStore caches with millions of ledgers.

### Initial Optimization Attempts (Commits 311f51f60 - 23862d0ed)
**SQL Batching**: First attempt batched SQL operations to reduce calls from ~3000 to ~5 per ledger. Dropped due to complex SQLite lock interactions.

**Cache Bypass**: Added `isPinned()` checks throughout the system - excluded pinned ledgers from AcceptedLedgerCache and LedgerHistory, created UNCACHED node types (later renamed pinnedACCOUNT_NODE/pinnedTRANSACTION_NODE). This addressed memory bloat but revealed a deeper issue.

### The Rotation Crisis (Commits da0f3d6e8 - 6c3e3c2a8)
Discovered the existing code was trying to **copy ALL pinned ledgers on EVERY rotation** - writing full trees each time. With millions of ledgers, the next rotation would be due before the copy finished.

**Evolution**: Removed broken code → Explored LVM snapshots with rotation templates → Breakthrough question: **Why tie pinned ledgers to rotation at all?**

### DatabasePinned: The Elegant Solution (Commit baca3b60a)
Dual-backend architecture: rotating backend (RWDB) for hot ledgers, persistent backend (NuDB/RocksDB) for pinned nodes. NodeObjectType routing handles everything - no copying, no complexity.

### The Parallelization Saga (Commits 03604584f - 1673c5999)
With memory and rotation solved, focus shifted to speed. Catalogue loading was taking 48 days - unacceptable.

**Attempts**:
- **Parallel save jobs** (highly concurrent) via JobQueue
- **Pointer-diff flushing** to identify changed nodes despite COW limitations
- **Canonicalize-then-flush** separating CPU work (hashing) from I/O (flushing)
- **JobQueueAdapter** for intelligent routing between standalone/production modes
- **Dedicated SQL threads** to isolate database operations

**The COW Problem**: SHAMap's copy-on-write uses binary ownership (cowid=0 or non-zero), can't track deltas in ledger chains. Snapshots cause all nodes to reset to cowid=0, breaking parallel flushing patterns.

**The Reality**: These optimizations provided marginal improvements but couldn't overcome the fundamental bottleneck - SQLite write performance degrades after 20-30 minutes regardless of tuning. Loading faster just created a long tail of pending SQL jobs that kept ledgers in memory.

### The Great Simplification (Commit 93a16ac3b)
After extensive experimentation: return to enlightened single-threading. 585 lines deleted.

**Realization**: Catalogue loading is a last resort tool. In live environments, optimizing it **steals resources from P2P and consensus**. Once custom offline tools exist, it likely won't be used.

**Final Approach**:
- Synchronous execution in standalone mode (`pendSaveValidated` with `sync=true`)
- Single-threaded, no JobQueue complexity
- Accept that it's slow - optimize for simplicity and reliability instead

### Final Cleanup (Commits d084ed393 - 8964c72ee)
Methodically removed all experiments: flushByPointerDiff, SHAMapCOW_test suite, SQL batching infrastructure, file splitting, startup cleanup, rotation commands, annotation markers. Restored jtPUBOLDLEDGER limit to battle-tested value of 2. Fixed assertion failures for pinned ledgers in standalone mode. Restored pinned range persistence.

**What Survived**:
- DatabasePinnedImp (dual backend architecture)
- Cache bypass for pinned ledgers
- Pinned-range-aware publishing and online_delete behavior
- Synchronous standalone execution
- Simple, reviewable code

## Commit History

Everything above describes the final design and current behavior. The history below is the exploration path, including approaches that were later removed or simplified.

| Commit | Files | Description | Notes | Action Items |
|--------|-------|-------------|-------|--------------|
| 311f51f60 | Node.cpp, jss.h, Catalogue.cpp | **Initial optimization attempt**: Batch SQL operations to reduce DB calls from ~3000 to ~5 per ledger. Combines DELETE operations, batches AccountTransactions and transaction metadata INSERTs. Adds diagnostic options to catalogue_load (`do_pinning`, `do_save_synchronous`) to debug memory issues by preventing JobQueue buildup when ledgers load faster than they can be saved. | Dropped due to unclear sqlite write lock interactions. Leaving existing, already battle-tested code | |
| 74ce7f5be | LedgerMaster.h/cpp, Node.cpp | **Cache bypass for pinned ledgers**: Adds `isPinned()` method for thread-safe checking. Modifies `storeLedger()` to skip LedgerHistory insertion when `pin=true` and adds locking for `mPinnedLedgers` access. Skips AcceptedLedgerCache population in `saveValidatedLedger()` for pinned ledgers. | Avoiding populating cache with pinned ledgers | |
| ec6b4446f | NodeObject.h, DatabaseNodeImp.h/cpp | **NodeStore cache bypass**: Adds `hotACCOUNT_NODE_UNCACHED` and `hotTRANSACTION_NODE_UNCACHED` types. Implements cache bypass logic in `DatabaseNodeImp::store()` - when these types are detected, they're converted to regular types but skip cache insertion. Allows pinned ledger nodes to go directly to backend storage without filling the node store cache. | Later renamed to pinnedACCOUNT_NODE and pinnedTRANSACTION_NODE | |
| d09b2262e | JobTypes.h | **Increase publish workers**: Increases `jtPUBOLDLEDGER` worker limit from 2 to 8 to address backlog of SQLite save jobs during bulk loading. Intended to improve throughput by allowing more concurrent ledger publishing. | Reverted to 2 to match battle-tested tuning and reduce testing/review complexity. SQLite write performance degrades inconsistently after 20-30 minutes despite initial improvements | |
| 23862d0ed | RPCCall.cpp, jss.h, Catalogue.cpp | **Diagnostic flags for memory debugging**: Adds `sync` flag (shorthand for synchronous saves), `no_db` flag (skip database saves entirely), and command-line parser support. Uses uncached node types when flushing and adds periodic NodeStore cache sweeping every 100 ledgers. Fixes journal usage. Groping around trying to determine why memory blowouts persisted despite cache changes - testing in production mode made it confusing to distinguish normal memory usage from catalogue-related issues. | Exploratory/diagnostic work. Later discovered node store cache itself was the issue - it's a vestigial 2014 workaround that ripple upstream is in process of removing | |
| eb811b12c | Catalogue_test.cpp | **Fix test race conditions after cache bypass**: Adds `getLedgerWithRetry()` helper that polls for ledger availability with configurable intervals and timeouts. Tests were breaking because cache bypass changes mean ledgers are only available after async publishAcqLedger jobs complete, not immediately. Uses struct with named parameters for readability. Requires loaded ledgers to be validated before comparison. | Largely intact in final version. Essential fix for testing with cache bypass architecture | |
| da0f3d6e8 | SHAMapStoreImp.cpp, DatabaseRotatingImp.cpp | **Rotation template via external command**: Removes broken `ensurePinnedLedgersInWritable()` code (was trying to copy all pinned ledgers on every rotation, writing full trees each time). Adds `online_delete_rotate_to_command` config to call external script that prepares new backend using LVM snapshots. Script receives pinned ledger ranges, returns path to prepared database. Idea: use append-only NUDB snapshots to make rotation cheap. | Tried to support template database for rotation with LVM snapshots. Later abandoned for DatabasePinnedImp architecture which uses separate persistent backend for pinned nodes | |
| 683078129 | (many files) | **Merge origin/dev**: Sync with upstream dev branch. Includes CMake refactoring, removal of Flatmap backend, dependency management changes. | Merge commit | |
| 6c3e3c2a8 | LedgerMaster.h/cpp, SHAMapStoreImp.h/cpp, State.h/cpp | **Persist pinned ranges across restarts**: Adds PinnedLedgers table to state.db with serialized range sets (e.g., "1000000-2000000,3000000-3500000"). Implements get/setPinnedRanges in SavedStateDB. Adds setPinnedLedgersRangeSet to LedgerMaster for restoration. Loads pinned ranges from database on startup. Pinned ranges were previously only tracked in RAM and lost on restart - loading is slow so persistence is essential. | Later: loading timing moved to when first validated ledger is loaded (rather than startup) | |
| 52f051cbe | (many files) | **Merge dev**: Richard merged dev branch. Same changes as 683078129. | Merge commit | |
| 03604584f | SHAMapStoreImp.h/cpp, Catalogue.cpp | **Naive parallelization attempt**: Implements Richard's "run-ahead" suggestion - parallel save jobs via JobQueue with MAX_CONCURRENT_SAVES=1000. Creates LedgerSaveJob structure with snapshots, moves flushDirty into parallel jobs. Batches pinned range persistence every 100 ledgers. Context: catalogue loading was taking 48 days, needed speed improvements. | Abandoned/simplified later. Multiple issues: (1) SQLite can't handle it sustainably (2) SHAMap COW implementation is not thread-safe for pattern of single thread building ledgers while background threads flush them | |
| b2ce9d625 | SHAMapStoreImp.cpp | **Add testing config flag**: Adds `load_complete_ledgers_pinned` config option to skip loading pinned ranges on startup. For testing/diagnostics - allows easy reset without manual database manipulation. | Removed/simplified later | |
| 78e02e15b | RPCCall.cpp, jss.h, Catalogue.cpp | **Simplify catalogue_load API**: Removes diagnostic flags (`do_save_synchronous`, `no_db`, `sync`) that were added for debugging memory issues. Renames `do_pinning` to `pin` with `no_pin` flag. Cleans up conditional checks - always save to database now. These flags were for testing various settings to isolate memory problems. | Further simplified later - only `ignore_hash` remains in final version | |
| ad80c6a90 | - | **Merge self**: Merging remote branch with itself. | Merge commit | |
| 2146b2dcd | LedgerMaster.cpp, NetworkOPs.cpp, Catalogue.cpp | **Fix memory bloat vs. publishing conflict**: Modifies `storeLedger` to skip mLedgerHistory when pinning (avoids memory bloat with millions of ledgers). Updates `findNewLedgersToPublish` to skip publishing pinned ledgers (except most recent) using RangeSet subtraction. Creates temporary JobQueue for parallel saves during catalogue load. Core issue: pinned ledgers weren't in mLedgerHistory (by design), but tryAdvance needed them there to publish. | Partially kept: RangeSet subtraction logic and storeLedger pinning remain. Temporary JobQueue replaced with pendSaveValidated + synchronous saves in standalone mode | |
| 345b0247c | LedgerMaster.cpp | **Documentation clarification**: Expands comments explaining why RangeSet subtraction is critical. Clarifies standalone mode is for production catalogue loading (not just tests) - loads large packs without consensus stealing resources. Without optimization, tryAdvance would publish ALL pinned ledgers, keeping them and SHAMap nodes in memory forever. | Documentation commit - comment simplified/rewritten later. Current comment is correct as-is (standalone mode is just for tests since in-memory databases only) | No action needed - current comment correct |
| c72e44e31 | LedgerMaster.cpp, SHAMapStoreImp.cpp, Node.cpp | **SQL batching with size limits**: Implements BulkSQLQueryBuilder to batch SQL operations while respecting SQLite's 1MB query size limit (identified by Richard during review). Uses conservative batch sizes (850KB for AccountTransactions, 500KB for Transactions). Fixes complete_ledgers to show pinned historical ranges via setFullLedger() hook - when mCompleteLedgers is empty and mPinnedLedgers isn't, assigns pinned ranges. Defaults load_complete_ledgers_pinned to false in standalone. | SQL batching dropped/reverted - SQLite sustained write performance not good enough, safety tradeoffs not worth it. Complete_ledgers fix KEPT (line 971-973 in LedgerMaster.cpp with detailed comments) | |
| f0fd3d237 | LedgerMaster.cpp | **Documentation**: Documents why pinned ledgers aren't available immediately until network synced, even though database is trusted on startup. The "theatre": mValidLedgerSeq/mValidLedgerSign start at 0, causing getValidatedLedgerAge() to return 2 WEEKS, making isValidatedOld() block ALL queries until first network validation. Database is trusted but RPC layer won't serve queries until node syncs. Documents 4 blocking conditions, complete_ledgers lifecycle, and why pinned ledgers excluded from publish. | Trying to make sense of startup blocking behavior - conclusion: safety mechanism forces network sync before serving queries despite having valid pinned ledgers in database | |
| 32218beb7 | LedgerMaster.cpp, SHAMapStoreImp.h/cpp, DatabaseCon.h, Catalogue.cpp | **Pinned ledger support for online_delete**: Updates `clearSql` (later `clearSqlRanges`) to accept pinnedLedgers parameter and subtract pinned ranges from deletion target (`target -= pinned`). Adds `SQLITE_FORCE_FILES` env var (standalone mode uses in-memory databases by default, this forces files). Prevents catalogue_load when online_delete configured. Fixes null pointer with isInitialRotation parameter. Rotation script stderr logging. HACK: deleteInterval=4 for testing. Documents standalone ledger 2 contamination issue. | clearSqlRanges with pinned parameter KEPT (actually uses it). SQLITE_FORCE_FILES removed (if gone, should remove standalone mode support except unittests). HACK removed. Rotation script support removed with DatabasePinnedImp migration | |
| 90dda527c | SHAMapStoreImp.cpp | **Replace test HACK with env var**: Removes hardcoded deleteInterval=4 override. Adds `OVERRIDE_ROTATE_INTERVAL` environment variable for testing rotation. Only allows overriding to smaller values than configured. Error handling for invalid values. | Removed later along with rotation script support | |
| 28c9c3635 | SHAMapStoreImp.h/cpp, SQLiteDatabase.h/cpp, RWDBDatabase.h, Node.h/cpp | **Range-based deletion for online_delete**: Replaces primitive `DELETE WHERE seq < X` with range-based deletion to support pinned ledgers. Adds `deleteInRange` methods with LIMIT parameter to SQLiteDatabase interface. Implements delete loop with batch size and sleep between operations. Skips empty ranges to avoid locks. Uses SQLite indexes for efficiency. `clearSqlRanges` calls `deleteInRanges(target)` lambda with RangeSet. | Went through iterations - some form of range-based deletion kept (deleteTransactionsInRange exists with minSeq/maxSeq/limit parameters). Core concept survives but implementation evolved | |
| 299da9b52 | SHAMapStoreImp.h/cpp, SQLiteDatabase.h/cpp, RWDBDatabase.h, impl/SQLiteDatabase.cpp | **Binary partitioning workaround**: SQLite needs SQLITE_ENABLE_UPDATE_DELETE_LIMIT compile flag for DELETE with LIMIT. Binary partitioning works around this: counts rows before delete (skips empty ranges), subdivides large sparse ranges to find data clusters, deletes manageable batches with sleeps, uses density thresholds to avoid excessive subdivision. Ensures online deletion respects database lock niceness without custom SQLite builds. | Binary partitioning approach REMOVED. Current Node.cpp uses simple subquery workaround: `DELETE FROM X WHERE rowid IN (SELECT rowid FROM X WHERE conditions LIMIT N)`. Caller (SHAMapStoreImp) loops with sleeps until deleted==0. Much simpler than binary partitioning | |
| c813f9ce9 | SHAMapStoreImp.cpp | **Cap deletion threshold**: Prevents deleting more than `deleteBatch_ * 10` rows in single operation even when data is dense. Ensures database locks held for reasonable durations. Dense ranges with >1000 rows subdivided through binary partitioning to maintain lock niceness. | Removed with binary partitioning. Current approach just uses deleteBatch_ limit directly in loop-until-zero pattern | |
| 410a10165 | SHAMapStoreImp.cpp | **Revert**: Reverts commit c813f9ce9 (cap deletion threshold). | Revert | |
| e6bc1fdb1 | SHAMapStoreImp.h/cpp, SQLiteDatabase.h/cpp, RWDBDatabase.h, impl/SQLiteDatabase.cpp | **Revert**: Reverts commit 299da9b52 (binary partitioning for online delete). | Revert - binary partitioning removed | |
| f86e456ab | impl/SQLiteDatabase.cpp | **Subquery workaround for DELETE LIMIT**: SQLite doesn't support DELETE...LIMIT without SQLITE_ENABLE_UPDATE_DELETE_LIMIT compile flag. Implements subquery pattern: `DELETE WHERE rowid IN (SELECT rowid ... LIMIT n)`. Allows batched deletion (100 rows at a time) to work with pinned ledgers without holding database locks too long. | KEPT - this is the simple solution that remains in Node.cpp today (lines 179-188) | |
| ee846b8a2 | Node.h, impl/Node.cpp, impl/SQLiteDatabase.cpp | **Refactor SQL to Node.cpp helpers**: Moves SQL delete range implementation from SQLiteDatabase.cpp to detail namespace helpers in Node.cpp where all SQL logic is centralized. Adds overloaded `deleteRange()` with optional limit parameter. Uses subquery pattern for DELETE with LIMIT. SQLiteDatabase now calls helpers instead of implementing SQL directly. | KEPT - Node.cpp still has this structure (deleteRange at lines 171-203). Clean separation of concerns | |
| 866748ab3 | SHAMapStoreImp.cpp | **Style**: Apply clang-format to all changes since f0fd3d237a. | Formatting | |
| accf75ed0 | SHAMapStoreImp.h/cpp | **Startup bulk cleanup**: Adds `performStartupCleanup()` that runs on startup to efficiently remove unpinned ledgers in bulk before node begins processing. Uses actual database max ledger as reference point. Keeps recent history (last deleteInterval ledgers) and pinned ranges. Performs bulk deletion without batching/sleeping for fast cleanup. Can be skipped with `SKIP_SHAMAPSTORE_STARTUP_CLEANUP` env var. Allows database to reach desired state immediately instead of waiting hours for incremental deletion. | REMOVED later - caused unexplained/confusing delays on startup. Could have been solved with logging but ultimately removed in final simplification before review | |
| bb78732d1 | SHAMapStoreImp.cpp | **Documentation**: Adds warning comments about potentially surprising behavior of startup cleanup feature. Notes aggressive bulk deletion happens immediately on startup, may be unexpected for users expecting gradual deletion. Should potentially be opt-in rather than opt-out. Lists TODO items for making feature less surprising. | Removed with performStartupCleanup feature | |
| 87130717a | Application.cpp, SHAMapStoreImp.h/cpp, Node.cpp, Config.cpp, Catalogue.cpp, Config_test.cpp | **Snapshot for spec source projection**: Adds `//@@start` and `//@@end` annotation markers for living documentation system (projected-source.md format). Markers like `shamap-store-init`, `database-choice-condition`, `shamap-store-rotate-with-lock`, etc. allow documentation to reference specific code sections with line numbers and GitHub URLs. | All markers stripped later (context summary mentions "Stripped all //@@start and //@@end annotation markers") | |
| baca3b60a | SHAMapStoreImp.cpp, Node.cpp, Config.cpp, DatabasePinnedImp.h/cpp | **THE BREAKTHROUGH: DatabasePinned implementation**: Implements dual-backend architecture that routes nodes based on NodeObjectType. pinnedACCOUNT_NODE/pinnedTRANSACTION_NODE/pinnedLEDGER → persistent backend (NuDB, never rotates). Regular hot types → memory backend (RWDB, rotates to prevent OOM). Inherits from DatabaseRotating for SHAMapStoreImp integration. Config validation requires type=rwdb, online_delete, pinned_type, pinned_path. This is the elegant solution that replaced rotation command/LVM snapshot complexity - separate backends for separate purposes. | KEPT - DatabasePinnedImp.h/cpp still exist and are core to the architecture. This is the solution that survived | |
| c94efa14b | CMake, Application.cpp, SHAMapStoreImp.h/cpp, DatabasePinnedImp.h/cpp, DatabaseRotatingImp.h, Catalogue.cpp, SHAMap.h/cpp | **Refine DatabasePinned + COW workaround**: Refines DatabasePinned implementation using composition with DatabaseRotatingImp. Fixes catalogue_load to flush nodes synchronously in main thread. WORKAROUND: Ripple's COW system uses binary ownership (cowid=0 or non-zero) instead of tracking deltas between versions. When building ledger chains, snapshots cause all nodes to get cowid=0, preventing background flushing. Solution: flush synchronously in main thread until COW can be redesigned for chain building. | KEPT - flushDirty calls remain in Catalogue.cpp (lines 1242-1243). COW limitation documented as reason for synchronous approach | |
| 9db5b823e | SHAMapStoreImp.cpp, Node.cpp, Config.cpp, DatabasePinnedImp.h/cpp | **Style**: clang-format. | Formatting | |
| 09a02210b | Catalogue.cpp, SHAMap.h/cpp | **Pointer-diff flushing experiment**: Implements `flushByPointerDiff()` using pointer comparison to identify modified nodes between parent/child maps. Works around COW's binary ownership (cowid=0/non-zero) which can't track deltas in ledger chains. Uses fact that unchanged subtrees share exact pointers and node store deduplicates by value making duplicate writes no-ops. Adds `flushMapsInMain` toggle for performance comparison. Initial tests showed marginal improvement, proving SHAMap needs fundamental redesign for ledger chains. | REMOVED - experiment didn't provide enough benefit. Synchronous flushDirty approach kept instead | |
| 5d03e9560 | SuiteJournal.h | **Test fix**: std::endl triggers flush() → sync() on shared log buffer. Multiple threads racing in sync() cause str()/str("") operations to corrupt buffer state, leading to crashes and double frees. Adds mutex to serialize access to suite.log, preventing concurrent sync() calls. | Test infrastructure fix - unrelated to pinned ledgers work | |
| 493fc4edd | SHAMapStoreImp.h, Catalogue.cpp, SHAMap.h/cpp, Catalogue_test.cpp | **Give up on background flushing**: Commit message says "do in main, give up for now". Sets flushMapsInMain=true, giving up on parallel/async flushing experiments. Removes debug cout statements, adds detailed trace logging. Expands comments explaining COW conflicts with deferred flushing: COW ID resets to 0 during snapshot chains, background threads can't flush because flushDirty() sees cowid=0. Solution: flush synchronously in main thread before snapshots. | Pragmatic decision - synchronous single-threaded approach wins due to SHAMap/COW design constraints | |
| 27f2702dd | CMake, Catalogue.cpp, Catalogue_test.cpp, SHAMapCOW_test.cpp | **Add SHAMapCOW test suite**: Creates comprehensive test file (882 lines) to explore and document SHAMap COW behavior. Refactors Catalogue.cpp (317 line changes). Tests likely explore the COW ID reset issue, snapshot chains, and flush behavior to understand why parallel flushing doesn't work. | Test file removed later - exploration/documentation effort to understand COW constraints | |
| 9cba41f59 | SHAMapStoreImp.cpp, Catalogue.cpp, Catalogue_test.cpp, SHAMapCOW_test.cpp | **Separate SQL operations into async jobs**: Queues SQL saves as separate jobs after flush completes to eliminate pauses. Adds comprehensive COW synchronization documentation. Disables NodeStore sweep during loading (causes stutters). Adds warnings about async startup cleanup causing lock contention. Documents discovered issues with background flushing and SQLite bottlenecks. Expands SHAMapCOW_test.cpp (925 lines). | Another attempt to optimize by separating concerns - likely simplified/removed later in final cleanup | |
| 7cf8b040b | SHAMapStoreImp.cpp, Node.cpp, Catalogue.cpp | **Documentation and status logging**: Adds WAL checkpoint performance documentation explaining 1000-ledger rhythm. Documents async startup cleanup causing confusing delays during catalogue load. Adds status logging every 100 ledgers in standalone mode with dedicated journal. Adds loadThreads and fileBytesProcessed fields to catalogue status (to be wired up). Documents JobQueue competition between catalogue temp queue and main app queue. Explains database contention patterns in standalone vs non-standalone modes. | Status logging removed in final simplification. Some documentation may remain | |
| c1d1cc76f | DBInit.h, Node.cpp, DatabaseCon.h/cpp, SociDB.cpp, Catalogue.cpp, SHAMapCOW_test.cpp | **[broken] scribble notes**: Commit message indicates broken/WIP state. Large refactor of Catalogue.cpp (486 line changes). Changes to database initialization and SOCI. Exploration/experimentation commit. | WIP/broken commit - exploration that likely gets cleaned up or reverted | |
| 9491c8ca2 | Catalogue.cpp, SHAMap.cpp, SHAMapCOW_test.cpp | **Canonicalize-then-flush approach**: "We've fixed the COW flushing issue!" Separates CPU work (hashing/canonicalization in main thread) from I/O work (flushing in background). Canonicalizes state map in main thread (makes nodes immutable, cowid=0), then uses flushByPointerDiff in background. TX hash verified after flush. Sets flushMapsInMain=false, canonicalizeInMain=true. Expands SHAMapCOW_test (579 lines added). This made things slightly faster, but ultimately doesn't solve the SQLite bottleneck. | Removed in final simplification - clever optimization fighting an inescapable bottleneck | |
| 1673c5999 | Catalogue.cpp | **Remove synchronization chain**: Canonicalize-then-flush makes synchronization unnecessary since unshare() makes nodes immutable (thread-safe). Removes mutexes, waiting, synchronization overhead. Allows true parallel I/O. Simplifies code: 119 lines deleted, 42 added. Claims "dramatically simplifies the code and improves performance" by eliminating synchronization and enabling parallel flushes. | Another optimization iteration - ultimately removed when reverting to synchronous approach | |
| ea1c4880e | CMake, Catalogue.cpp → Catalogue.h/cpp + CatalogueCreate.cpp + CatalogueLoad.cpp + CatalogueStatus.cpp | **Split monolithic file**: Splits large Catalogue.cpp (~2000 lines) into separate handler files for maintainability: shared utilities in Catalogue.cpp/h, doCatalogueCreate in CatalogueCreate.cpp, doCatalogueLoad in CatalogueLoad.cpp, doCatalogueStatus in CatalogueStatus.cpp. Fixes duplicate CMake entries. File was getting long in the tooth. | Reverted later in interest of review simplicity - easier to review one file than multiple | |
| 8b0aa6734 | Catalogue.h/cpp, CatalogueCreate.cpp, CatalogueLoad.cpp | **JobQueueAdapter for intelligent routing**: Creates JobQueueAdapter class that routes jobs based on environment. Standalone: creates own JobQueue with aggressive thread counts for max parallelism. Production: routes through app's JobQueue with polite priorities. Defines CatalogueJobType enum (FLUSH_SAVE, SQL_SAVE). Renames original CatalogueJobType to CatalogueStatusJobType. Adds comprehensive 3-queue architecture documentation. Enables catalogue to be aggressive in standalone while polite in production. | Removed - reverted to simple synchronous approach and file re-merge | |
| d3ce546a7 | CatalogueLoad.cpp | **Unify job routing through JobQueueAdapter**: Removes dedicated SQL thread and queue. Routes SQL saves through JobQueueAdapter with CatalogueJobType::SQL_SAVE. Maintains SQL metrics tracking. Simplifies cleanup by removing thread management. Uses consistent job routing for both flush and SQL operations. Claims "Performance remains excellent with 1140+ ledgers/sec throughput." Net: 142 lines deleted, 82 added. | Removed with JobQueueAdapter - all reverted to synchronous approach | |
| aeb2ff2cd | Catalogue.h, CatalogueLoad.cpp | **Commit nonsense before reverting**: Commit message says "commit nonsense before reverting to single thread". Large refactor: 192 lines added to Catalogue.h, 327 line changes to CatalogueLoad.cpp. About to give up on parallelization and go back to single-threaded approach. | WIP commit before major simplification - the moment of surrender to simplicity | |
| 93a16ac3b | JobTypes.h, CatalogueLoad.cpp | **THE GREAT SIMPLIFICATION**: "After extensive experimentation with parallel processing, return to enlightened single-threading." Single-threaded execution avoids fighting SHAMap's COW architecture. SQL saves queued through JobQueue (jtPUBOLDLEDGER) to yield to P2P operations. Direct execution in standalone mode (no JobQueue overhead). Removes all JobQueueAdapter and parallel processing infrastructure. "This 'good enough' performance is appropriate for occasional range expansion on live servers." Net: 585 lines deleted, 293 added. Embraces simplicity. | KEPT - this is the core approach that survived. Wisdom over cleverness | |
| 67e7db56a | Node.cpp, JobTypes.h, CatalogueLoad.cpp | **Add metrics and tweak limits**: Adds metrics tracking time spent waiting for SQL job promises - shows 67% of time spent idle waiting for JobQueue slots. Increases jtPUBOLDLEDGER concurrent limit from 2 to 4 to reduce contention with P2P ledger acquisition. Re-enables size-based batching in Node.cpp (MAX_ENTRIES_PER_BATCH = 0). | Reverted - jtPUBOLDLEDGER back to 2 (battle-tested tuning). Metrics removed in simplification | |
| 637d2bb9e | DBInit.h, RWDBDatabase.h, SQLiteDatabase.h, Node.h/cpp, impl/SQLiteDatabase.cpp, CatalogueLoad.cpp | **Batch commit experiment**: Large changes to Node.cpp (366 lines changed), CatalogueLoad.cpp (486 line changes). Adds batching infrastructure to database layer. Another attempt to optimize SQL writes through batching. Net: 635 additions, 316 deletions. | Reverted - goes back to the SQL batching experiments that were already proven problematic in commit 311f51f60 | |
| 63d53020a | DBInit.h, RWDBDatabase.h, SQLiteDatabase.h, Node.h/cpp, impl/SQLiteDatabase.cpp, CatalogueLoad.cpp | **Revert batch commit experiment**: Immediately reverts previous commit. Net: 635 deletions, 316 additions (exact opposite). Lesson re-learned. | Revert | |
| 3d706b695 | DBInit.h, Node.cpp | **Restore conservative settings**: Reverts DBInit.h to original WAL settings (1.5MB journal_size_limit). Maintains conservative MAX_ENTRIES_PER_BATCH=1 (proven behavior). Simplifies BulkSQLQueryBuilder logic. Removes ill-conceived BATCH_DELAY that would sleep while holding write lock. Updates comments clarifying entire ledger save is in single transaction. Net: 123 deletions, 30 additions. Back to known-good configuration. | Simplification - removing experiments, returning to proven settings | |
| 5a09d1bc9 | Node.cpp | **Restore original SQL batching**: Separates DELETE statements for SOCI compatibility. Restores multi-VALUES grouping per transaction for AccountTransactions. Adds entryLimit parameter to BulkSQLQueryBuilder for future tuning. Sets conservative defaults matching original behavior (1 entry/statement). Fixes memory reservation to use reasonable sizes (150-2048 bytes). | Cleanup - restoring original proven SQL patterns | |
| d049e73ba | Node.cpp | **Remove SQL batching infrastructure**: Removes BulkSQLQueryBuilder class (200+ lines of unused complexity). Keeps deleteRange functions for online_delete support. Keeps pinned ledger routing (pinnedLEDGER vs hotLEDGER). Keeps selective AcceptedLedger caching for memory efficiency. Keeps exception-based error handling. Removes per-transaction DELETE. Net: 229 deletions, 32 additions. "Simplifies code review while preserving all necessary functionality for pinned ledger ranges and online_delete integration." | Major cleanup - removing all SQL batching complexity from commit 311f51f60 | |
| 697376785 | CMake, Ledger.h/cpp, RPCCall.cpp, Catalogue.cpp (merge back split files) | **Merge split files + synchronous standalone**: Merges Catalogue.h/cpp + CatalogueCreate.cpp + CatalogueLoad.cpp + CatalogueStatus.cpp back into single Catalogue.cpp. Removes split files (commit ea1c4880e reversed). Uses synchronous save for standalone mode - pendSaveValidated executes synchronously in standalone. Net: 2248 deletions, 1138 additions (big net reduction from removing all experiments). | Major cleanup - reversing file split for review simplicity, implementing sync=standalone pattern | |
| d084ed393 | SHAMap.h/cpp | **Remove flushByPointerDiff complexity**: Reverts SHAMap to simpler implementation without pointer-diff optimization (from commit 09a02210b). "The complexity wasn't necessary for the current approach and the standard flushDirty method is sufficient." Net: 251 deletions, 2 additions. | Cleanup - removing canonicalize-then-flush/pointer-diff experiments | |
| 1df1e2875 | CMake, Application.cpp, DatabaseCon.h/cpp, SociDB.cpp, DatabaseRotatingImp.cpp, jss.h, SHAMapCOW_test.cpp | **Remove debugging remnants**: Removes lambda wrappers from Application.cpp initialization (m_shaMapStore, m_nodeStore, nodeFamily_) - were functionally identical to direct initialization. Removes entire SHAMapCOW_test.cpp (1698 lines) - exploration/documentation test file from commits 27f2702dd and 9cba41f59. Removes SociDB.cpp (85 lines). Cleans up jss.h. Net: 1834 deletions, 15 additions. | Major cleanup - removing test suite and debugging code | |
| e0236fc7b | LedgerMaster.cpp | **Fix assertion failure for pinned ledgers**: Prevents assertion `alpAccepted->getLedger().get() == lpAccepted.get()` when publishing pinned ledgers. Issue occurs in standalone mode when catalogue_load switches to newer ledger, triggering tryAdvance → findNewLedgersToPublish → pubLedger for pinned ledgers that were never cached (to save memory). Keeps pinned ledger optimization in findNewLedgersToPublish to skip publishing intermediate pinned ledgers (except most recent). Updates comments explaining AcceptedLedger cache issue. | KEPT - critical fix for standalone mode, remains in current codebase (mentioned in context summary) | |
| fc1270845 | SHAMapStoreImp.h/cpp, Config.cpp, Config_test.cpp | **Simplify SHAMapStore + remove annotations**: Removes performStartupCleanup (commit accf75ed0) that caused async lock contention and confusing delays. Removes rotation_to_command logic from makeBackendRotating (commit da0f3d6e8). Removes isInitialRotation parameter. Removes boost/process include. Strips all //@@start and //@@end annotation markers (commit 87130717a). Net: 273 deletions, 12 additions. Clean code ready for review. | Major cleanup - removing startup cleanup, rotation command support, and projected-source annotations | |
| c618d330f | JobTypes.h | **Restore jtPUBOLDLEDGER limit to 2**: Reverts increase from 2 to 4 (commit 67e7db56a). Original limit of 2 is sufficient and battle-tested. | Revert - back to proven tuning | |
| fa94fa66b | Catalogue_test.cpp | **Fix test ordering and reliability**: Restores original test function ordering to minimize diff noise. Adds getLedgerWithRetry helper for reliable ledger polling (from commit eb811b12c). Enhances test logging and error handling. Fixes race conditions in ledger availability checks. | Test improvements - KEPT (getLedgerWithRetry mentioned in context summary as remaining) | |
| 8964c72ee | LedgerMaster.cpp, SHAMapStore.h, SHAMapStoreImp.h/cpp, SQLiteDatabase.h, Node.cpp, Catalogue.cpp | **Restore pinned range persistence + simplify**: Adds setPinnedRanges to SHAMapStore interface. Calls setPinnedRanges from Catalogue.cpp after every ledger (single row UPDATE). Removes OVERRIDE_ROTATE_INTERVAL env var. Removes checkHasExtents documentation/comments. Updates LedgerMaster comment removing "caller must manually insert" confusion. Removes per-transaction DELETE comment. | KEPT - pinned range persistence restored (mentioned in context summary) | |
| 173835c6c | (many files) | **Merge origin/dev**: Sync with upstream dev branch. | Merge commit | |
| 412be9a99 | Config.cpp, DatabaseNodeImp.cpp, Config_test.cpp | **Remove noisy logging + fix DatabasePinned validation**: Removes info-level logging for pinnedACCOUNT_NODE/pinnedTRANSACTION_NODE stores in DatabaseNodeImp (too noisy). Fixes DatabasePinned config validation - just requires `type` is set, not specifically rwdb (can be any backend: NuDB, RocksDB, etc). Fixes Config_test.cpp formatting. | KEPT - logging cleanup and config fix mentioned in context summary | |
| 39d1c4390 | conanfile.py | **Upgrade OpenSSL**: Upgrades from 1.1.1u to 3.6.0 (#587). | FROM DEV | |
| 094f01100 | SetHook_test.cpp | **Fix test name**: Fixes testcase name from "Test float_emit" to "Test emit" (#580). | FROM DEV | |
| e580f7cfc | .vscode/settings.json | **Enable format on save**: Sets editor.formatOnSave to true (#578). | FROM DEV | |
| ad0531ad6 | RWDBDatabase.h, Handshake.cpp | **Fix warnings**: Fixes compiler warnings - getJson asString(), network ID handling (#509). | FROM DEV | |
| 00dcfcc15 | (many files) | **Merge origin/dev**: Sync with upstream dev branch. | Merge commit | |
| 1f12b9ec5 | Multiple logging files, CMakeLists, conanfile | **Enhanced logging with file:line**: Adds -DBEAST_ENHANCED_LOGGING with file:line numbers for JLOG macro (#552). | FROM DEV | |
| 15c7ad6f7 | SetHook.cpp, SetSignerList.cpp, Feature.h, Feature.cpp, TxFlags.h | **Fix invalid tx flags**: Adds fixInvalidTxFlags amendment to validate flags in SetHook and SetSignerList transactions (#514). | FROM DEV | |
| 83f09fd8a | (many files) | **Merge origin/dev**: Sync with upstream dev branch. | Merge commit | |
| 687ccf420 | MultiSign_test.cpp | **Remove unused variable**: Removes unused `enabled` variable (#592). | FROM DEV | |
| 9c8b00540 | applyHook.cpp | **Improve preflight logging**: Better logging for transaction preflight failures in applyHook (#566). | FROM DEV | |
| ebd1289a5 | DatabasePinnedImp.h | **Simplify comment**: Removes "memory" qualifier from rotation comment for clarity. | Comment cleanup - pinned ledgers work | |
| db7a2be7d | (many files) | **Merge origin/dev**: Final merge with upstream dev branch. | Merge commit (HEAD) | |


src/ripple/app/rdb/backend/detail/impl/Node.cpp
```diff
@@ -266,76 +271,130 @@ saveValidatedLedger(
...
+                }
+
+                metadataBulk += acceptedLedgerTx->getTxn()->getMetaSQL(
+                    seq, acceptedLedgerTx->getEscMeta());
+            }
+
+            // Execute bulk inserts - this is where the performance gain
+            // happens. Instead of N individual INSERT statements per
+            // transaction type, we execute just 2 bulk statements for the
+            // entire ledger.
```
RichardAH: Isn't the max length of a query set to 1Mib by default? What happens when the query is larger?

src/ripple/rpc/handlers/Catalogue.cpp
```diff
@@ -924,7 +926,29 @@ doCatalogueLoad(RPC::JsonContext& context)
...
-    JLOG(context.j.info()) << "Opening catalogue file: " << filepath;
+    bool do_pinning = true;
+    bool do_save_synchronous = false;  // Default: asynchronous saves
+    bool no_db = false;                // Default: save to database
+
+    // Parse diagnostic options if provided
+    if (context.params.isMember(jss::do_pinning))
+        do_pinning = context.params[jss::do_pinning].asBool();
+
+    if (context.params.isMember(jss::do_save_synchronous))
```
RichardAH: the `do_` is inconsistent with other parameters. I think `pin` and `async` would be fine

sublimator: These were diagnostic nonsense for debugging, I was going to remove them, 
It's still in draft

src/ripple/app/rdb/backend/detail/impl/Node.cpp
sublimator: good question

src/ripple/rpc/handlers/Catalogue.cpp
sublimator: I removed them but left "pin" in

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1380,51 +1399,79 @@ LedgerMaster::findNewLedgersToPublish(
...
 
+    // Create a range of ledgers we need to publish
+    RangeSet<std::uint32_t> toPublish;
+    toPublish.insert(range(pubSeq, valSeq));
+
+    // IMPORTANT: This RangeSet subtraction is critical for standalone mode when
+    // loading large catalogue files (millions of ledgers). Standalone mode is
+    // NOT just for tests - it's used for loading catalogue packs without
+    // consensus stealing resources. Without this optimization, tryAdvance would
+    // attempt to publish ALL pinned ledgers, keeping them and their SHAMap
```
sublimator: TODO: standalone mode would need to be configured to use actual database files for standalone mode to actually work

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2926,6 +2926,7 @@ NetworkOPsImp::pubLedger(std::shared_ptr<ReadView const> const& lpAccepted)
         app_.getAcceptedLedgerCache().fetch(lpAccepted->info().hash);
     if (!alpAccepted)
     {
+        std::cout << "is this happening?? " << std::endl;
```
sublimator: TODO

src/ripple/app/ledger/impl/LedgerMaster.cpp
sublimator: But then you might get some weird ledgers circa the genesis ledger added to the sqlitedb

sublimator: Unless catalogue_create created its own db handles in standalone mode ... 
Which sounds complicated

sublimator: Now we have DatabasePinned, we should just NOT support standalone mode, because it uses in-memory databases, and because the JobQueue is limited

Builds/CMake/RippledCore.cmake
```diff
@@ -668,7 +669,6 @@ target_sources (rippled PRIVATE
   src/ripple/rpc/handlers/ValidatorListSites.cpp
   src/ripple/rpc/handlers/Validators.cpp
   src/ripple/rpc/handlers/WalletPropose.cpp
-  src/ripple/rpc/handlers/Catalogue.cpp
```
sublimator: there was a duplicate, so this is not actually deleting it from the list, but removing the duplicate

src/ripple/core/impl/Config.cpp
```diff
@@ -1017,6 +1017,37 @@ Config::loadFromString(std::string const& fileContents)
...
+    if (!RUN_STANDALONE)
+    {
+        auto db_section = section(ConfigSection::nodeDatabase());
+        if (db_section.exists("pinned_type"))
+        {
+            // Ensure base type is rwdb
+            if (auto type = get(db_section, "type", ""); type != "rwdb")
+            {
+                Throw<std::runtime_error>(
+                    "pinned_type requires type=rwdb for memory backend");
```
sublimator: TODO: technically, this could work with nudb as well

src/ripple/nodestore/impl/DatabaseNodeImp.cpp
```diff
@@ -33,9 +33,26 @@ DatabaseNodeImp::store(
 {
     storeStats(1, data.size());
 
+    // Check if this is an uncached type
+    bool skipCache = false;
+    if (type == pinnedACCOUNT_NODE)
+    {
+        type = hotACCOUNT_NODE;
+        skipCache = true;
+        JLOG(j_.info()) << "Storing uncached ACCOUNT_NODE, hash: " << hash;
```
sublimator: TODO: remove the logs!

src/xrpld/app/misc/SHAMapStoreImp.cpp
```diff
@@ -118,6 +136,9 @@ SHAMapStoreImp::SHAMapStoreImp(
 
     get_if_exists(section, "online_delete", deleteInterval_);
 
+    // Always initialize state database for pinned ranges persistence
```
sublimator, line 139: TODO: 

Is this actually appropriate for standalone mode? 
Unittests? 
Where is it made?

sublimator: State Database in Standalone/Unit Test Mode

  The Flow:

  1. Config.cpp:396 - In standalone mode, if no database_path is configured, dataDir.clear() leaves it empty ""
  2. DatabaseCon.cpp:113-117 - Empty database_path is allowed in standalone (only throws if NOT standalone)
  3. SociDB.cpp:68 - path = config.legacy("database_path") → ""
  4. SociDB.cpp:53-56 - getSociSqliteInit("state", "", ".db"):
  boost::filesystem::path file("");        // empty path
  if (is_directory(file))                  // false
      file /= name + ext;                  // SKIPPED
  return file.string();                    // returns ""
  5. SociDB.cpp:94 - s.open(soci::sqlite3, "") → In-memory SQLite database

  Result: The state.db database (which stores pinned ranges) is in-memory in standalone mode.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -206,19 +228,57 @@ SHAMapStoreImp::makeNodeStore(int readThreads)
...
-        fdRequired_ += dbr->fdRequired();
-        dbRotating_ = dbr.get();
-        db.reset(dynamic_cast<NodeStore::Database*>(dbr.release()));
+        // Check if DatabasePinned should be created
+        if (nscfg.exists("pinned_type"))
+        {
+            // DatabasePinned uses the same rotation backends as
+            // DatabaseRotating but adds a persistent backend for pinned nodes
+
+            // Create persistent backend (NuDB) for pinned data
```
sublimator: TODO: could be ANY supported type of backend, not just NuDB

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -550,54 +660,47 @@ SHAMapStoreImp::makeBackendRotating(std::string path)
...
-        if (min < lastRotated)
-            std::this_thread::sleep_for(backOff_);
-        if (healthWait() == stopping)
-            return;
-    }
-    JLOG(journal_.debug()) << "finished deleting from: " << TableName;
+    JLOG(journal_.debug()) << "Pruning " << tableName
+                           << ". Target ranges: " << to_string(target);
+
+    // Process each interval with binary partitioning
```
sublimator: Old straggler comment

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -550,54 +660,47 @@ SHAMapStoreImp::makeBackendRotating(std::string path)
...
-SHAMapStoreImp::clearSql(
+SHAMapStoreImp::clearSqlRanges(
     LedgerIndex lastRotated,
-    std::string const& TableName,
+    RangeSet<std::uint32_t> const& pinned,
+    std::string const& tableName,
     std::function<std::optional<LedgerIndex>()> const& getMinSeq,
-    std::function<void(LedgerIndex)> const& deleteBeforeSeq)
+    std::function<void(RangeSet<std::uint32_t> const&)> const& deleteInRanges,
+    std::optional<RangeSet<std::uint32_t>> const& complete)
```
sublimator: complete is never used

src/xrpld/app/ledger/detail/LedgerMaster.cpp
```diff
@@ -1753,6 +1813,76 @@ LedgerMaster::getPinnedLedgersRangeSet()
...
+    {
+        Throw<std::runtime_error>(
+            "Expected mPinnedLedgers to be empty on startup");
+    }
+    mPinnedLedgers.assign(range_set);
+
+    // Pinned ledgers are by definition complete (they were fully saved and are
+    // retrievable from the database). However, we DON'T add them to
+    // mCompleteLedgers here at startup. Instead, we wait for the first ledger
+    // validation in setFullLedger() to do a one-time copy. This approach:
```
sublimator: TODO: assumption is that complete ledgers never goes backward for any reason

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -279,6 +339,54 @@ SHAMapStoreImp::copyNode(std::uint64_t& nodeCount, SHAMapTreeNode const& node)
...
+            JLOG(journal_.warn())
+                << "Failed to parse pinned ranges: " << rangesStr;
+        }
+    }
+}
+
+// NOTE: Startup cleanup of unpinned ledgers was removed.
+// The async cleanup caused confusing lock contention and performance issues,
+// especially when users tried operations like catalogue_load immediately after
+// startup. The cleanup would hold SQLite write locks for extended periods,
```
sublimator: TODO: rehome comment

src/ripple/app/rdb/backend/impl/SQLiteDatabase.cpp
```diff
@@ -98,6 +99,24 @@ class SQLiteDatabaseImp final : public SQLiteDatabase
     void
     deleteAccountTransactionsBeforeLedgerSeq(LedgerIndex ledgerSeq) override;
 
+    std::size_t
+    deleteLedgersInRange(
```
sublimator: These are redundant, and in the header?

sublimator: This pattern of redundancy was actually pre-existing

src/ripple/core/impl/Config.cpp
```diff
@@ -1017,6 +1017,37 @@ Config::loadFromString(std::string const& fileContents)
...
+            {
+                Throw<std::runtime_error>(
+                    "pinned_path is required when pinned_type is set");
+            }
+
+            // Ensure online_delete is set for rotation
+            if (auto delete_interval = get(db_section, "online_delete", 0);
+                delete_interval == 0)
+            {
+                Throw<std::runtime_error>(
```
sublimator, line 1122: It's not for memory, and it's more that it's pointless to use pinning without online_delete

src/xrpld/nodestore/detail/DatabaseNodeImp.cpp
```diff
@@ -33,9 +33,24 @@ DatabaseNodeImp::store(
 {
     storeStats(1, data.size());
 
+    // Check if this is an uncached type
+    bool skipCache = false;
```
sublimator: Does the DatabasePinneImp.cpp need this as well, or this class is part of its composition ?

src/xrpld/nodestore/detail/DatabasePinnedImp.cpp
```diff
@@ -0,0 +1,196 @@
...
+                << "Hot stores: " << count << " (type=" << type << ")";
+        }
+        rotating_.store(type, std::move(data), hash, ledgerSeq);
+    }
+}
+
+std::shared_ptr<NodeObject>
+DatabasePinnedImp::fetchNodeObject(
+    uint256 const& hash,
+    std::uint32_t ledgerSeq,
```
sublimator, line 97: Is there any opportunity to use the ledgerSeq here intelligently to predict the relevant backend ?

src/xrpld/nodestore/detail/DatabaseNodeImp.cpp
sublimator: No, DatabasePinnedImp routes pinnedXXX directly to a Backend instance which does NOT cache. 

OR 


No, DatabasePinnedImp doesn't need that logic.

  - persistent_ is a raw storage Backend (NuDB/RocksDB/etc.) with no cache
  - Backend::store() just writes the NodeObject directly to disk
  - Pinned types stay as pinnedACCOUNT_NODE (no conversion needed)
  - DatabaseNodeImp needs the conversion because it has a cache to skip
  - DatabasePinnedImp has no cache - it just routes by type to different backends

src/xrpld/app/ledger/detail/LedgerMaster.cpp
sublimator: Well, it CAN go backwards:

```

  The mCompleteLedgers Backward Question

  IT CAN GO BACKWARD! Found the answer to your earlier TODO:

  - Line 622: clearLedger(seq) - Erases individual sequences from mCompleteLedgers
  - Line 905, 928: fixMismatch() - Calls clearLedger() when finding mismatches
  - Line 1010: failedSave() - Calls clearLedger() on save failures
  - Line 2023: getLedgerBySeq() - Calls clearLedger() when ledger not found
  - Lines 2072-2093: clearPriorLedgers() - Mass erases ranges then re-adds pinned ones
```
 
TOOD: 
clearLedger should probably not be able to clear a pinned ledger

sublimator: Net result is correct ✅

  - clearLedger(seq) - Already checks, won't touch pinned ledgers
  - clearPriorLedgers(range) - Erases everything, then re-adds pinned ones back
  - Both end up with pinned ledgers still in mCompleteLedgers
  - Lock held throughout
  
  All good

src/ripple/core/impl/Config.cpp
```diff
@@ -1017,6 +1017,38 @@ Config::loadFromString(std::string const& fileContents)
             }
         }
     }
+
+    // Add DatabasePinned validation
+    if (!RUN_STANDALONE)
```
sublimator: TODO: 

Why not validate when in standalone mode?

src/ripple/app/rdb/backend/detail/impl/Node.cpp
sublimator: reverted to original tested queries

src/xrpld/app/ledger/detail/LedgerMaster.cpp
```diff
@@ -525,6 +537,20 @@ LedgerMaster::clearLedger(std::uint32_t seq)
     mCompleteLedgers.erase(seq);
 }
 
+bool
+LedgerMaster::isPinned(std::uint32_t seq)
+{
+    std::lock_guard sl(mCompleteLock);
```
RichardAH: should we be using read-write locks?

sublimator: iirc this was using existing conventions with mCompleteLock 

Let me investigate and see if this can be improved because we may be adding a lot more contention than prior 

We may also consider a completely separate lock for the pinned state members too?

src/xrpld/app/rdb/backend/detail/Node.cpp
```diff
@@ -157,6 +157,52 @@ deleteBeforeLedgerSeq(
...
+void
+deleteRange(
+    soci::session& session,
+    TableType type,
+    LedgerIndex minSeq,
+    LedgerIndex maxSeq)
+{
+    session << "DELETE FROM " << to_string(type)
+            << " WHERE LedgerSeq >= " << minSeq
+            << " AND LedgerSeq <= " << maxSeq << ";";
```
RichardAH: sanity check and limit please

src/xrpld/app/rdb/backend/detail/Node.cpp
```diff
@@ -157,6 +157,52 @@ deleteBeforeLedgerSeq(
...
+}
+
+std::size_t
+deleteRange(
+    soci::session& session,
+    TableType type,
+    LedgerIndex minSeq,
+    LedgerIndex maxSeq,
+    std::optional<std::size_t> limit)
```
RichardAH: sanity check and limit here as well... if limit > (maxSeq - minSeq) then set it accordingly

sublimator: lemme check, 

iirc the limit was referring to actual rows which doesn't really have a relationship to maxSeq/minSeq which are referring to ledgers

sublimator: yeah, it's rows, I renamed to rowLimit and added some XRPL_ASSERT sanity checks it's empty or at least 1

src/xrpld/app/ledger/detail/LedgerMaster.cpp
```diff
@@ -524,7 +524,8 @@ LedgerMaster::haveLedger(std::uint32_t seq)
 void
 LedgerMaster::clearLedger(std::uint32_t seq)
 {
-    std::lock_guard sl(mCompleteLock);
+    std::lock_guard cl(mCompleteLock);
+    std::lock_guard pl(mPinnedLock);
```
RichardAH: this is deadlock central, prefer something like std::scoped_lock lock(mCompleteLock, mPinnedLock);

PR #549, opened 2025-07-23
UDP Super Highway [DO NOT MERGE]

RichardAH:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -739,10 +740,20 @@ TxQ::apply(
...
+
+    view.getAndResetKeysTouched();
     if (auto directApplied = tryDirectApply(app, view, tx, flags, j))
+    {
+        app.getHashRouter().setTouchedKeys(
+            transactionID, view.getAndResetKeysTouched());
         return *directApplied;
+    }
+
+    return {telCAN_NOT_QUEUE, false};
```
shortthefomo, line 756: Unconditional early return in TxQ::apply breaks transaction queueing for all non-replay networks
After tryDirectApply returns an empty optional (fee too low for direct apply), TxQ::apply unconditionally returns {telCAN_NOT_QUEUE, false} at line 756. This makes all subsequent code — the entire transaction queue logic — unreachable dead code.


The added line return {telCAN_NOT_QUEUE, false}; at TxQ.cpp:756 is placed after the tryDirectApply block but before the queueing logic. For non-replay networks (NETWORK_ID != 65534), when a transaction's fee is too low for direct application, tryDirectApply returns {} and the function hits this unconditional return.
This means:
Transactions that previously would be queued (returning terQUEUED) now always fail with telCAN_NOT_QUEUE
The replay network queue path at lines 772-810 is also dead code (though replay networks bypass this via tryDirectApply always succeeding)
The standard queue logic at lines 812+ including fee averaging, account queue management, and all replacement logic is completely unreachable
Impact: Transaction queueing is completely broken for all non-replay networks. Any transaction that doesn't qualify for immediate direct application will be rejected instead of queued.

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -197,43 +176,164 @@ doSubmit(RPC::JsonContext& context)
...
+                errorResult[jss::error] = "invalidTransaction";
+                errorResult[jss::error_exception] =
+                    "tx_blobs element must be string";
+                errorPromise.set_value(std::move(errorResult));
+                futures.push_back(errorPromise.get_future());
+                continue;
+            }
+
+            const std::string txBlobStr = txBlobs[i].asString();
+            futures.push_back(std::async(
```
shortthefomo, line 276: Batch submit via std::async captures RPC context by reference causing data races
The doSubmit batch transaction handler launches std::async tasks that capture context by reference ([&context, ...]). Multiple threads concurrently access the non-thread-safe RPC::JsonContext struct.

src/ripple/ledger/impl/OpenView.cpp
```diff
@@ -263,6 +263,9 @@ OpenView::rawTxInsert(
     std::shared_ptr<Serializer const> const& txn,
     std::shared_ptr<Serializer const> const& metaData)
 {
+    if (txExists(key))
```
shortthefomo, line 266: LogicError function returns despite [[noreturn]] attribute — undefined behavior
The detail::accessViolation() call in LogicError was commented out, causing the [[noreturn]] function to fall off the end and return.

src/ripple/app/misc/HashRouter.cpp
```diff
@@ -128,4 +128,20 @@ HashRouter::shouldRelay(uint256 const& key)
...
 }
 
+void
+HashRouter::setTouchedKeys(uint256 const& id, std::set<uint256>&& k)
+{
+    std::unique_lock<std::shared_mutex> lock(touchedKeysMutex_);
+    touchedKeysMap_.insert_or_assign(id, std::move(k));
+}
+
+std::optional<std::reference_wrapper<const std::set<uint256>>>
```
shortthefomo, line 138: getTouchedKeys returns reference that dangles after lock release
The HashRouter::getTouchedKeys function returns a std::reference_wrapper to data inside touchedKeysMap_, but the shared_lock protecting that data is released when the function returns.

PR #559, opened 2025-07-30
fix: upgrade CI to GCC 13 and fix compilation issues, fixes #557

### LINKED ISSUE #557: GCC 13 build failures: missing includes and explicit constructor requirements
## Fix
Add `#include <cstdint>` after line 4 in src/ripple/app/hook/Enum.h
Change {{}} to std::nullopt at line 324 in src/test/app/SetHook_test.cpp

## Environment
- GCC 13.3.0 (Ubuntu 24.04)
- CI uses GCC 11 (no issue there)

sublimator:
Updates CI workflow to use GCC 13 and fixes compiler issues

## Changes
- Upgraded CI workflow from gcc-11/g++-11 to gcc-13/g++-13
  - Release builder covers gcc 11
- Added missing `#include <cstdint>` header in `src/ripple/app/hook/Enum.h`
- Fixed issue in `SetHook_test.cpp` by explicitly specifying `std::vector<Json::Value>` type
- Removed overly broad cache restore keys that were causing cache pollution between different compiler versions
- Parameterized compiler version in workflow matrix
- Bumped cache version to force clean rebuild

Fixes #557


src/test/app/SetHook_test.cpp
```diff
@@ -321,7 +321,7 @@ class SetHook0_test : public beast::unit_test::suite
             HSFEE,
             ter(temMALFORMED));
 
-        env(ripple::test::jtx::hook(alice, {{}}, 0),
+        env(ripple::test::jtx::hook(alice, std::nullopt, 0),
```
sublimator, line 324: testing 123

RichardAH: as discussed in mm pass empty vec here

PR #561, opened 2025-08-02
Batch [DO NOT MERGE]

tequdev:
### Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [X] Tests (you added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation update
- [ ] Release

### API Impact


src/xrpld/app/hook/detail/applyHook.cpp
```diff
@@ -570,6 +570,9 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
         case ttREMARKS_SET: {
             break;
         }
+        case ttBATCH: {
```
tequdev, line 576: To be discussed:
1. Is Batch allowed as EmittedTxn?
2. If allowed, how should Callback be handled?

RichardAH: I vote yes and singularly at the end of the batch txn execution like any other txn. We might need additional hook apis to support the extra metadata

tequdev: Should `cbak()` handle both OuterTxn and InnerTxns individually?

Also, since InnerTxn doesn't have information on whether it was emitted from a Hook itself, we might need to set the `EmitDetails` field on InnerTxn as well.

tequdev:
New Hook API Ideas
- Get InnerTxn from OuterTxn/other InnerTx (Something like `otxn_batch_inner_slot(innerIndex, newSlotno)`?)
- Get OuterTxn from InnerTxn (Something like `otxn_batch_outer_slot(newSlotno)`?)

tequdev:
For now, we'll only allow BatchTx.InnerTransactions.sfAccount = HookAccount (since signatures from other accounts can't be provided), but in the future, we might consider including transactions from other accounts in InnerTxns if there's authorization via OtxnAccount.

tequdev:
Points for consideration:
Should we apply the HookCanEmit check to InnerTxns? (We probably should.)

PR #570, opened 2025-08-18
Fix rwdb memory leak with online_delete and remove flatmap

sublimator:


src/test/app/SHAMapStore_test.cpp
```diff
@@ -517,12 +521,96 @@ class SHAMapStore_test : public beast::unit_test::suite
...
+        cfg->section(ConfigSection::nodeDatabase()).set("online_delete", "256");
+        cfg->setupControl(
+            true, true, false);  // bQuiet, bSilent, bStandalone=false
+        return cfg;
+    }
+
+    void
+    testRWDBOnlineDeleteEnforcement()
+    {
+        testcase("RWDB online_delete enforcement");
```
sublimator: I don't really like this here, but it is what it is 
It's testing code in SHAMapStoreImp.cpp, which had preexisting validation code that throwed exceptions
Take the L and move on?

src/test/app/SHAMapStore_test.cpp
```diff
@@ -517,12 +521,96 @@ class SHAMapStore_test : public beast::unit_test::suite
...
+        // Test 1: RWDB without online_delete in standalone mode (should
+        // succeed)
+        {
+            jtx::Env env{*this, rwdbNoDeleteDefaultConfig()};
+            pass();
+        }
+
+        // Test 2: RWDB without online_delete NOT in standalone mode (should
+        // throw)
+        try
```
sublimator: If the component instantiation order changes, these tests could start failing. They are a bit iffy/brittle

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -118,6 +118,19 @@ SHAMapStoreImp::SHAMapStoreImp(
 
     get_if_exists(section, "online_delete", deleteInterval_);
 
+    // RWDB (in-memory backend) requires online_delete to prevent OOM
+    // Exception: standalone mode (used by tests) doesn't need online_delete
```
sublimator: todo, move to ripple::Config::loadFromString

src/ripple/app/rdb/backend/RWDBDatabase.h
```diff
@@ -463,11 +457,9 @@ class RWDBDatabase : public SQLiteDatabase
         for (const auto& [_, accountData] : accountTxMap_)
         {
             size += sizeof(AccountID) + sizeof(AccountTxData);
-            size += accountData.transactions.size() * sizeof(AccountTx);
-            for (const auto& [_, innerMap] : accountData.ledgerTxMap)
+            for (const auto& [_, txVector] : accountData.ledgerTxMap)
             {
-                size += sizeof(uint32_t) +
-                    innerMap.size() * (sizeof(uint32_t) + sizeof(size_t));
+                size += sizeof(uint32_t) + txVector.size() * sizeof(AccountTx);
```
RichardAH: is this right?
`    using AccountTx =
        std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;`
shouldn't we iterate txVector and find the byte sizes of the txns and meta?

same above for the size of ledgers

src/test/rdb/RelationalDatabase_test.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+        // Min sequences should remain the same (first transaction ledger)
+        auto newTxMinSeq = sqliteDb->getTransactionsMinLedgerSeq();
+        auto newAcctTxMinSeq = sqliteDb->getAccountTransactionsMinLedgerSeq();
+        BEAST_EXPECT(newTxMinSeq == txMinSeq);
+        BEAST_EXPECT(newAcctTxMinSeq == acctTxMinSeq);
+    }
+
+    void
+    run() override
```
RichardAH, line 728: should these tests be run against multiple relational database backends?

PR #573, opened 2025-08-20
Fix `slot` to return `bigint` when specified true for second argument

### LINKED ISSUE #466: slot tries to pack int64 into a JS Number
```
    if (argc == 2 && !!JS_ToBool(ctx, argv[1]))
        JS_FreeValue(ctx, *out);
        returnJS(data_as_int64(ptr, len));
```

I don't believe all int64 will pack into a JS Number

tequdev:

fix #466

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/hook/impl/applyHook.cpp
```diff
@@ -3432,7 +3433,10 @@ DEFINE_JS_FUNCTION(JSValue, slot, JSValue raw_slot_no)
     if (argc == 2 && !!JS_ToBool(ctx, argv[1]))
     {
         JS_FreeValue(ctx, *out);
-        returnJS(data_as_int64(ptr, len));
+        int64_t data = data_as_int64(ptr, len);
```
sublimator, line 3436: We definitely want a signed integer here, and not unsigned ? 
Is there any argument for unsigned?

WASM slot returns a signed integer, so it's inline with that at least

sublimator: This also leads to a discussion about JS vs WASM api convention adherence. The reason for existing is to make things easier.

Could change the api to `slot(n, 'uint64')` and `slot(n, 'int64')` ?

sublimator: Only tangentially related, I wonder why we aren't using JS_CLASS_UINT8_ARRAY, i.e. a typed array, as using an array of floats is a bit inefficient? 

https://github.com/Xahau/xahaud/issues/574

sublimator (approved):

PR #585, opened 2025-09-22
build: migrate to conan 2

sublimator:
# PR #585: Migrate to Conan 2

## Summary

This PR migrates the xahaud build system from Conan 1 to Conan 2, addressing deprecation warnings and preparing for Conan 1's end-of-life. The migration maintains full compatibility with both local development and Holy Build Box (HBB) release builds while simplifying several aspects of the build configuration.

#### Updated Command Syntax
- **Export commands**: Changed from `conan export <path> <reference>` to `conan export <path> --version <ver> --user <user> --channel <channel>`
- **Profile management**: Migrated from Conan 1 profile format to Conan 2's cleaner syntax
- **Cache paths**: Updated from `/cache/conan` to `/cache/conan2` with separate download and source caches
- **Version overrides**: Ensure consistency with `override=True` for boost/1.86.0, sqlite3/3.42.0, snappy/1.1.10

#### Profile Structure (Conan 2 Format)
```ini
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

[conf]
# Build tools configuration
tools.build:compiler_executables={"c": "/usr/bin/gcc", "cpp": "/usr/bin/g++"}
```

### 2. Dual Boost Configuration Documentation

The PR clarifies and documents the intentional dual Boost setup in HBB builds:

- **Manual Boost (minimal)**: Built in Docker for WasmEdge only (filesystem + system components)
- **Conan Boost (full)**: Provides all components needed by the application and dependencies (nudb, soci)

Both are version 1.86.0, ensuring symbol compatibility. The static linking resolves any potential conflicts at link time.

**Why not just use Conan's Boost for everything?** WasmEdge is compiled during the Docker image build (for caching efficiency), while Conan runs later inside a container using that image. The Docker-cached WasmEdge build needs Boost libraries to be present at Docker build time, before Conan has even run.

**Why not manage everything through Conan?** This is technically possible but would require:
- A custom WasmEdge recipe that builds from source (upstream only provides prebuilt binaries)
- A custom LLVM 14.x recipe (Conan Center jumped from 13.x to 19.x, skipping the version WasmEdge needs)
- Figuring out how to make Conan recipes use system libraries (not straightforward in Conan's isolated build environment)

The current approach of Docker-cached manual builds + Conan for standard dependencies is a pragmatic compromise that avoids maintaining complex custom recipes.

### 3. Build Tool Requirements

Added `tool_requires()` for build tools to ensure proper glibc compatibility:
- `protobuf/3.21.12` - Provides protoc compiler
- `grpc/1.50.1` - Provides gRPC plugins
- `b2/5.3.2` (HBB only) - Forces source build to avoid GLIBCXX version issues

Additionally, the HBB Docker configuration includes:
```ini
[conf]
# Force building from source for packages with binary compatibility issues
*:tools.system.package_manager:mode=build
```
This ensures all Conan packages are built from source with HBB's gcc-toolset-11, avoiding prebuilt binaries that might have incompatible glibc requirements.

### 4. Updated Custom WasmEdge Recipe

Updated `external/wasmedge/conanfile.py` to handle prebuilt WasmEdge binaries:
- Updated `package_id()` implementation for Conan 2

### 5. CI/CD Updates

All GitHub Actions workflows updated for Conan 2:
- **Linux**: Full Conan 2 profile configuration
- **macOS**: CMake 3.31.7 for compatibility, Apple Clang workarounds
- **Docker**: Comprehensive dependency installation with proper caching

### HBB Build Environment

The Holy Build Box environment requires special handling for glibc 2.28 compatibility:
- Prebuilt b2 binaries require GLIBCXX_3.4.29 (too new for HBB's glibc 2.28)
- Solution: Force b2 to build from source via `tool_requires('b2/5.3.2')`
- Manual LLVM 14 and WasmEdge installation for static linking

**Key constraint**: WasmEdge requires LLVM 14.x specifically for WebAssembly compilation, but Conan Center only offers 11.x, 12.x, 13.x, or 19.x. This gap necessitates manual installation.

### Build Verification

Added comprehensive build verification:
- Verbose final link step to debug symbol resolution
- Full `ldd` output to verify dynamic dependencies
- `libcheck` validation for system library compliance

### Temporary File Handling

Improved temp file management in containerized builds:
```bash
export TMPDIR=/io/tmp
export TEMP=/io/tmp
export TMP=/io/tmp
```

## Migration Benefits

1. **Future-proof**: Conan 1 reaches end-of-life soon
2. **Cleaner syntax**: Conan 2's profile and configuration format is more maintainable
3. **Better caching**: Separate download/source caches improve build performance
4. **Explicit dependencies**: `tool_requires()` clearly separates build tools from library dependencies
5. **Improved documentation**: Dual Boost setup is now clearly explained

## Testing

- ✅ Local development builds (macOS, Linux)
- ✅ HBB release builds with glibc 2.28 compatibility
- ✅ CI/CD pipelines (all platforms)
- ✅ Static linking verification via `ldd` and `libcheck`

## Breaking Changes

None. The migration maintains full backwards compatibility:
- Same compiler requirements
- Same dependency versions (except minor bumps for bug fixes)
- Same build outputs
- Same binary compatibility

## Notes for Reviewers

1. The dual Boost setup in HBB is intentional and documented - both versions are 1.86.0
2. The `tool_requires('b2/5.3.2')` in HBB is necessary to avoid glibc incompatibility
3. The WasmEdge recipe handles prebuilt binaries only (no source build option from upstream)

## Checklist

- [x] All tests pass
- [x] HBB release builds work
- [x] Backwards compatibility maintained

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -23,21 +23,26 @@ class Xrpl(ConanFile):
         'unity': [True, False],
     }
 
+    import os
+
+    # Check if we're in HBB release build environment
+    IS_HBB_BUILD = os.getenv('HBB_RELEASE_BUILD') == '1'
+
     requires = [
-        'boost/1.86.0',
         'date/3.0.1',
         'libarchive/3.6.0',
-        'lz4/1.9.3',
+        'lz4/1.9.4',
         'grpc/1.50.1',
         'nudb/2.0.8',
         'openssl/1.1.1u',
-        'protobuf/3.21.9',
-        'snappy/1.1.10@xahaud/stable',
+        'protobuf/3.21.12',
         'soci/4.0.3@xahaud/stable',
-        'sqlite3/3.42.0',
-        'zlib/1.2.13',
-        'wasmedge/0.11.2',
+        'zlib/1.3.1',
     ]
+
+    # Only include deps that aren't manually installed in HBB
+    if not IS_HBB_BUILD:
+        requires.append('wasmedge/0.11.2@xahaud/stable')
 
     default_options = {
         'assertions': False,
@@ -51,41 +56,41 @@ class Xrpl(ConanFile):
         'tests': True,
         'unity': False,
 
-        'cassandra-cpp-driver:shared': False,
-        'date:header_only': True,
-        'grpc:shared': False,
-        'grpc:secure': True,
-        'libarchive:shared': False,
-        'libarchive:with_acl': False,
-        'libarchive:with_bzip2': False,
-        'libarchive:with_cng': False,
-        'libarchive:with_expat': False,
-        'libarchive:with_iconv': False,
-        'libarchive:with_libxml2': False,
-        'libarchive:with_lz4': True,
-        'libarchive:with_lzma': False,
-        'libarchive:with_lzo': False,
-        'libarchive:with_nettle': False,
-        'libarchive:with_openssl': False,
-        'libarchive:with_pcreposix': False,
-        'libarchive:with_xattr': False,
-        'libarchive:with_zlib': False,
-        'libpq:shared': False,
-        'lz4:shared': False,
-        'openssl:shared': False,
-        'protobuf:shared': False,
-        'protobuf:with_zlib': True,
-        'rocksdb:enable_sse': False,
-        'rocksdb:lite': False,
-        'rocksdb:shared': False,
-        'rocksdb:use_rtti': True,
-        'rocksdb:with_jemalloc': False,
-        'rocksdb:with_lz4': True,
-        'rocksdb:with_snappy': True,
-        'snappy:shared': False,
-        'soci:shared': False,
-        'soci:with_sqlite3': True,
-        'soci:with_boost': True,
+        'cassandra-cpp-driver/*:shared': False,
+        'date/*:header_only': True,
+        'grpc/*:shared': False,
+        'grpc/*:secure': True,
+        'libarchive/*:shared': False,
+        'libarchive/*:with_acl': False,
+        'libarchive/*:with_bzip2': False,
+        'libarchive/*:with_cng': False,
+        'libarchive/*:with_expat': False,
+        'libarchive/*:with_iconv': False,
+        'libarchive/*:with_libxml2': False,
+        'libarchive/*:with_lz4': True,
+        'libarchive/*:with_lzma': False,
+        'libarchive/*:with_lzo': False,
+        'libarchive/*:with_nettle': False,
+        'libarchive/*:with_openssl': False,
+        'libarchive/*:with_pcreposix': False,
+        'libarchive/*:with_xattr': False,
+        'libarchive/*:with_zlib': False,
+        'libpq/*:shared': False,
+        'lz4/*:shared': False,
+        'openssl/*:shared': False,
+        'protobuf/*:shared': False,
+        'protobuf/*:with_zlib': True,
+        'rocksdb/*:enable_sse': False,
+        'rocksdb/*:lite': False,
+        'rocksdb/*:shared': False,
+        'rocksdb/*:use_rtti': True,
+        'rocksdb/*:with_jemalloc': False,
+        'rocksdb/*:with_lz4': True,
+        'rocksdb/*:with_snappy': True,
+        'snappy/*:shared': False,
+        'soci/*:shared': False,
+        'soci/*:with_sqlite3': True,
+        'soci/*:with_boost': True,
     }
 
     def set_version(self):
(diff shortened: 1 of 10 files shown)


dangell7:
We will need to update the BUILD.md file as well. Conan 1 commands don't work for Conan 2.

dangell7:
Can you update wasmedge to solve those other issues? 

We didn't do wasm like that for caching we did it like that because it was just easier to install it and link then to build from source in cmake. When we first were developing we installed wasm locally and linked it there was no cake build. So I don't think there's any reason not to just use Conan for everything now, if you can.

sublimator:
> We didn't do wasm like that for caching 

I might not have been clear there. "WasmEdge is compiled during the Docker image build (for caching efficiency". What that was trying to say is that it's done in the (cached) Docker image build step. If it were done inside the "release-builder.sh" script it would be slow. Conan installed boost is not available at that point. So that's why it's not trivial to simply use conan boost to compile wasmedge. 

Your next question might be: "So why not just compile boost AND Wasm edge using Conan?" 

There's no recipe available for wasmedge to actually build it from source. It has dependencies on llvm and the llvm linker and boost. LLVM itself, at version 14, is not available on conan. I do not know about the linker. 

> So I don't think there's any reason not to just use Conan for everything now, if you can.

This is not news though. It's what was happening in the Conan 1 build. We're only really special casing things for the HBB build, and we're using the Conan wasmedge recipe. 

We CAN use conan for everything (nothing is impossible), but we'll need to make custom recipes for wasmedge that don't just download prebuilt binaries. We'll also have to manage the dependencies that we're currently installing in HBB, including llvm and the linker.

> Key constraint: WasmEdge requires LLVM 14.x specifically for WebAssembly compilation, but Conan Center only offers 11.x, 12.x, 13.x, or 19.x. This gap necessitates manual installation.

As per ^ 

I didn't really think it was worth it.

dangell7:
> > We didn't do wasm like that for caching
> 
> I might not have been clear there. "WasmEdge is compiled during the Docker image build (for caching efficiency". What that was trying to say is that it's done in the (cached) Docker image build step. If it were done inside the "release-builder.sh" script it would be slow. Conan installed boost is not available at that point. So that's why it's not trivial to simply use conan boost to compile wasmedge.
> 
> Your next question might be: "So why not just compile boost AND Wasm edge using Conan?"
> 
> There's no recipe available for wasmedge to actually build it from source. It has dependencies on llvm and the llvm linker and boost. LLVM itself, at version 14, is not available on conan. I do not know about the linker.
> 
> > So I don't think there's any reason not to just use Conan for everything now, if you can.
> 
> This is not news though. It's what was happening in the Conan 1 build. We're only really special casing things for the HBB build, and we're using the Conan wasmedge recipe.
> 
> We CAN use conan for everything (nothing is impossible), but we'll need to make custom recipes for wasmedge that don't just download prebuilt binaries. We'll also have to manage the dependencies that we're currently installing in HBB, including llvm and the linker.
> 
> > Key constraint: WasmEdge requires LLVM 14.x specifically for WebAssembly compilation, but Conan Center only offers 11.x, 12.x, 13.x, or 19.x. This gap necessitates manual installation.
> 
> As per ^
> 
> I didn't really think it was worth it.

This recipe has a llvm dependency? https://conan.io/center/recipes/wasmedge?version=0.11.2? Yeah maybe its HBB but I've built xahau using this recipe before. 

But ok doing it like this is fine.

dangell7 (dismissed):

sublimator:
> This recipe has a llvm dependency? https://conan.io/center/recipes/wasmedge?version=0.11.2? > Yeah maybe its HBB but I've built xahau using this recipe before.

Yeah, that works outside of HBB just fine, but all it does is download pre-built binaries (which aren't suitable for HBB, being shared libs that also link to other shared libs)

sublimator:
> We will need to update the BUILD.md file as well. Conan 1 commands don't work for Conan 2.

On it !

release-builder.sh
```diff
@@ -90,29 +92,37 @@ RUN /hbb_exe/activate-exec bash -c "dnf install -y epel-release && \
         llvm14-static llvm14-devel && \
     dnf clean all"
 
-# Install Conan and CMake
-RUN /hbb_exe/activate-exec pip3 install "conan==1.66.0" && \
+# Install Conan 2 and CMake
+RUN /hbb_exe/activate-exec pip3 install "conan>=2.0" && \
```
tequdev: In this specify, wouldn't conan3 be installed when conan3 is released??

conanfile.py
```diff
@@ -23,22 +23,27 @@ class Xrpl(ConanFile):
         'unity': [True, False],
     }
 
+    import os
+
+    # Check if we're in HBB release build environment
+    IS_HBB_BUILD = os.getenv('HBB_RELEASE_BUILD') == '1'
```
tequdev: It may be clearer to simply use the variable name `IS_STATIC_BUILD`.

HBB is a tool for static builds.

release-builder.sh
sublimator: Oooh, ooops, thanks !

conanfile.py
sublimator: I would argue IS_STATIC_BUILD would be too generic and misleading. It's specifically for changing the requirements when building for HBB.

It also controls workarounds to build some components from source for older glibc versions, using tool_requires.

It's simple and you know exactly what it's used for. As you know, the HBB build messes around with overriding cmake modules and doing sed surgeries. 

Simply setting HBB_RELEASE_BUILD -> IS_STATIC_BUILD would not be enough, so I think the existing name is actually more appropriate, thanks.

tequdev: I didn't like that the name of a tool only used in release-builder was being used here. However, it's fine as is.

sublimator: Yeah, I feel ya, the conanfile knowing about some concrete build environment is kind of strange, but it's self-documenting vs having to send a bunch of more abstract yet "pure" parameters that we'd likely only ever really use in HBB anyway. 

At the end of the day, it's a python file for a reason, and it's not THAT much of a stretch from checking other environment context like the operating system etc

sublimator: I'll have a quick look at some alternatives while I'm here anyway

sublimator: Had a look at the alternatives. Here are the options:

**Current approach (env var):**
```python
IS_HBB_BUILD = os.getenv('HBB_RELEASE_BUILD') == '1'

def requirements(self):
    if not IS_HBB_BUILD:
        requires.append('wasmedge/0.11.2@xahaud/stable')

def build_requirements(self):
    if IS_HBB_BUILD:
        self.tool_requires('b2/5.3.2')
```
Usage: `export HBB_RELEASE_BUILD=1`

**Alternative 1 (Conan option - unified):**
```python
options = {'hbb_mode': [True, False]}
default_options = {'hbb_mode': False}

def requirements(self):
    if not self.options.hbb_mode:
        self.requires('wasmedge/0.11.2@xahaud/stable')

def build_requirements(self):
    if self.options.hbb_mode:
        self.tool_requires('b2/5.3.2')
```
Usage: `conan install .. -o hbb_mode=True`

**Alternative 2 (separate options):**
```python
options = {
    'with_wasmedge': [True, False],
    'tool_requires_b2': [True, False]
}

def requirements(self):
    if self.options.with_wasmedge:
        self.requires('wasmedge/0.11.2@xahaud/stable')

def build_requirements(self):
    if self.options.tool_requires_b2:
        self.tool_requires('b2/5.3.2')
```
Usage: `conan install .. -o with_wasmedge=False -o tool_requires_b2=True`

What do you think?

Let me look at #2

sublimator: Option 2 impl in [c470ab2](https://github.com/Xahau/xahaud/pull/585/commits/c470ab2169e107aa996443c40d7518d9fc9d037a)

tequdev (approved):

PR #590, opened 2025-10-14
Cron (on ledger cronjobs)

RichardAH:
This feature allows each Account to set up an `ltCRON` object (using txn `SetCron`) which contains a delay and repeat count.

When the delay is satisfied the validators attempt to insert a pseudo `ttCron` into the end of the ledger (up to 128 of these.) This is achieved by iterating sequential keys in the range of valid Cron keylets.

A Cron keylet has an 8 byte namespace prefix, followed by 4 bytes of BE timestamp (when the Cron is to be executed) followed by 20 bytes of Account ID (for whom it is to be executed.)

When a `ttCRON` is applied, it self-deletes the `ltCRON` object that spawned it. The `sfOwner` of the `ltCRON` object becomes the `sfOwner` in the `ttCRON` and is a weak transactional stakeholder.

This allows hooks to request a wakeup call up to 1 year in advance and up to 256 times before needing to call `SetCron` again.

To remove an existing Cron submit a SetCron with both DelaySeconds and RepeatCount absent. Submitting a txn with only DelaySeconds implies a 0 repeat count, which is allowed for one-off transactions. A delay of 0 is also allowed and means apply as soon as it can be (in the next ledger typically).

The validators will always iterate from the oldest cron toward the end of the cron namespace, executing 128 per ledger until the next available cron (if any) exceeds the current last closed ledger timestamp.

### DIFF
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1477,6 +1477,63 @@ TxQ::accept(Application& app, OpenView& view)
         }
     }
 
+    // Inject cron transactions, if any
+    if (view.rules().enabled(featureCron))
+    {
+        uint32_t currentTime =
+            view.parentCloseTime().time_since_epoch().count();
+        uint256 klStart = keylet::cron(0, AccountID(beast::zero)).key;
+        uint256 const klEnd =
+            keylet::cron(currentTime + 1, AccountID(beast::zero)).key;
+
+        std::set<AccountID> cronAccs;
+
+        auto counter = 0;
+        while (++counter < 128 && klStart < klEnd)
+        {
+            std::optional<uint256 const> next = view.succ(klStart, klEnd);
+            if (!next.has_value())
+                break;
+
+            Keylet kl{ltANY, *next};
+
+            if (view.exists(kl))
+            {
+                auto sle = view.read(kl);
+                if (safe_cast<TxType>(sle->getFieldU16(sfLedgerEntryType)) ==
+                    ltCRON)
+                {
+                    // valid cron object, add it to the list
+                    cronAccs.emplace(sle->getAccountID(sfOwner));
+                }
+            }
+
+            klStart = *next;
+        }
+
+        auto const seq = view.info().seq;
+
+        // insert Cron pseudos for each of the accs we need to ping
+        for (AccountID const& id : cronAccs)
+        {
+            STTx cronTx(ttCRON, [=](auto& obj) {
+                obj[sfAccount] = AccountID();
+                obj[sfLedgerSequence] = seq;
+                obj[sfOwner] = id;
+            });
+
+            uint256 txID = cronTx.getTransactionID();
+
+            auto s = std::make_shared<ripple::Serializer>();
+            cronTx.add(*s);
+
+            app.getHashRouter().setFlags(txID, SF_PRIVATE2);
+            app.getHashRouter().setFlags(txID, SF_EMITTED);
+            view.rawTxInsert(txID, std::move(s), nullptr);
+            ledgerChanged = true;
+        }
+    }
+
     // Inject emitted transactions if any
     if (view.rules().enabled(featureHooks))
         do
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1489,7 +1489,8 @@ TxQ::accept(Application& app, OpenView& view)
         std::set<AccountID> cronAccs;
 
         auto counter = 0;
-        while (++counter < 128 && klStart < klEnd)
+        // include max 128 cron txns in the ledger
+        while (++counter < 129 && klStart < klEnd)
         {
             std::optional<uint256 const> next = view.succ(klStart, klEnd);
             if (!next.has_value())
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1500,8 +1500,8 @@ TxQ::accept(Application& app, OpenView& view)
             if (view.exists(kl))
             {
                 auto sle = view.read(kl);
-                if (safe_cast<TxType>(sle->getFieldU16(sfLedgerEntryType)) ==
-                    ltCRON)
+                if (safe_cast<LedgerEntryType>(
+                        sle->getFieldU16(sfLedgerEntryType)) == ltCRON)
                 {
                     // valid cron object, add it to the list
                     cronAccs.emplace(sle->getAccountID(sfOwner));
--- a/src/ripple/app/tx/impl/Cron.cpp
+++ b/src/ripple/app/tx/impl/Cron.cpp
@@ -173,6 +173,12 @@ Cron::doApply()
     return tesSUCCESS;
 }
 
+void
+Cron::preCompute()
+{
+    assert(account_ == beast::zero);
+}
+
 XRPAmount
 Cron::calculateBaseFee(ReadView const& view, STTx const& tx)
 {
--- a/src/ripple/app/tx/impl/Cron.cpp
+++ b/src/ripple/app/tx/impl/Cron.cpp
@@ -36,7 +36,6 @@ Cron::makeTxConsequences(PreflightContext const& ctx)
     return TxConsequences{ctx.tx, TxConsequences::normal};
 }
 
-
 NotTEC
 Cron::preflight(PreflightContext const& ctx)
 {
@@ -76,9 +75,8 @@ Cron::preflight(PreflightContext const& ctx)
         return temBAD_SEQUENCE;
     }
 
-    return tesSUCCESS;   
+    return tesSUCCESS;
 }
-
 
 TER
 Cron::preclaim(PreclaimContext const& ctx)
@@ -112,7 +110,7 @@ Cron::doApply()
     }
 
     uint256 ptr = sle->getFieldH256(sfCron);
-    Keylet klOld {ltCRON, ptr};
+    Keylet klOld{ltCRON, ptr};
     auto sleCron = view.peek(klOld);
     if (!sleCron)
     {
@@ -122,7 +120,7 @@ Cron::doApply()
 
     uint32_t delay = sleCron->getFieldU32(sfDelaySeconds);
     uint32_t recur = sleCron->getFieldU32(sfRepeatCount);
-    
+
     uint32_t currentTime = view.parentCloseTime().time_since_epoch().count();
 
     // do all this sanity checking before we modify the ledger...
@@ -134,11 +132,12 @@ Cron::doApply()
     // if there are further crons to do then a new one is created at the next
     // time point
 
-    if (!view.dirRemove(keylet::ownerDir(id), (*sleCron)[sfOwnerNode], klOld, false))
+    if (!view.dirRemove(
+            keylet::ownerDir(id), (*sleCron)[sfOwnerNode], klOld, false))
         return tefBAD_LEDGER;
 
     view.erase(sleCron);
-    
+
     if (recur == 0)
     {
         // already at last execution, stop here
@@ -151,9 +150,11 @@ Cron::doApply()
     // more executions remain, so create a new object
 
     Keylet klCron = keylet::cron(afterTime, id);
-    
-    // insert into owner dir, we don't need to check reserve because we've just deleted an object
-    auto const page = view.dirInsert(keylet::ownerDir(id), klCron, describeOwnerDir(id));
+
+    // insert into owner dir, we don't need to check reserve because we've just
+    // deleted an object
+    auto const page =
+        view.dirInsert(keylet::ownerDir(id), klCron, describeOwnerDir(id));
     if (!page)
         return tecDIR_FULL;
 
@@ -164,7 +165,7 @@ Cron::doApply()
     sleCron->setAccountID(sfOwner, id);
 
     sle->setFieldH256(sfCron, klCron.key);
-    
+
     view.insert(sleCron);
     view.update(sle);
 
--- a/src/ripple/app/tx/impl/Cron.cpp
+++ b/src/ripple/app/tx/impl/Cron.cpp
@@ -160,6 +160,7 @@ Cron::doApply()
 
     sleCron = std::make_shared<SLE>(klCron);
 
+    sleCron->setFieldU64(sfOwnerNode, *page);
     sleCron->setFieldU32(sfDelaySeconds, delay);
     sleCron->setFieldU32(sfRepeatCount, recur - 1);
     sleCron->setAccountID(sfOwner, id);
--- a/src/ripple/app/tx/impl/Cron.cpp
+++ b/src/ripple/app/tx/impl/Cron.cpp
@@ -1,0 +1,180 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2024 XRPL-Labs
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/tx/impl/Cron.h>
+#include <ripple/basics/Log.h>
+#include <ripple/core/Config.h>
+#include <ripple/ledger/View.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Indexes.h>
+#include <ripple/protocol/PublicKey.h>
+#include <ripple/protocol/Quality.h>
+#include <ripple/protocol/TxFlags.h>
+#include <ripple/protocol/st.h>
+
+namespace ripple {
+
+TxConsequences
+Cron::makeTxConsequences(PreflightContext const& ctx)
+{
+    return TxConsequences{ctx.tx, TxConsequences::normal};
+}
+
+
+NotTEC
+Cron::preflight(PreflightContext const& ctx)
+{
+    if (!ctx.rules.enabled(featureCron))
+        return temDISABLED;
+
+    auto const ret = preflight0(ctx);
+    if (!isTesSuccess(ret))
+        return ret;
+
+    auto account = ctx.tx.getAccountID(sfAccount);
+    if (account != beast::zero)
+    {
+        JLOG(ctx.j.warn()) << "Cron: Bad source id";
+        return temBAD_SRC_ACCOUNT;
+    }
+
+    // No point in going any further if the transaction fee is malformed.
+    auto const fee = ctx.tx.getFieldAmount(sfFee);
+    if (!fee.native() || fee != beast::zero)
+    {
+        JLOG(ctx.j.warn()) << "Cron: invalid fee";
+        return temBAD_FEE;
+    }
+
+    if (!ctx.tx.getSigningPubKey().empty() || !ctx.tx.getSignature().empty() ||
+        ctx.tx.isFieldPresent(sfSigners))
+    {
+        JLOG(ctx.j.warn()) << "Cron: Bad signature";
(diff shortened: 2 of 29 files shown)


src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,255 @@
...
+        Keylet klOld{ltCRON, sle->getFieldH256(sfCron)};
+
+        auto sleCron = view.peek(klOld);
+        if (!sleCron)
+        {
+            JLOG(j_.warn()) << "SetCron: Cron object didn't exist.";
+            return tefBAD_LEDGER;
+        }
+
+        if (safe_cast<TxType>(sleCron->getFieldU16(sfLedgerEntryType)) !=
```
sublimator: TxType ?

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1477,6 +1477,63 @@ TxQ::accept(Application& app, OpenView& view)
...
+            std::optional<uint256 const> next = view.succ(klStart, klEnd);
+            if (!next.has_value())
+                break;
+
+            Keylet kl{ltANY, *next};
+
+            if (view.exists(kl))
+            {
+                auto sle = view.read(kl);
+                if (safe_cast<TxType>(sle->getFieldU16(sfLedgerEntryType)) ==
```
sublimator: TxType?

src/ripple/app/tx/impl/Cron.cpp
```diff
@@ -0,0 +1,181 @@
...
+    auto const page =
+        view.dirInsert(keylet::ownerDir(id), klCron, describeOwnerDir(id));
+    if (!page)
+        return tecDIR_FULL;
+
+    sleCron = std::make_shared<SLE>(klCron);
+
+    sleCron->setFieldU32(sfDelaySeconds, delay);
+    sleCron->setFieldU32(sfRepeatCount, recur - 1);
+    sleCron->setAccountID(sfOwner, id);
```
sublimator, line 166: Needs sfOwnerNode set? as referenced in line 136?

src/ripple/protocol/impl/Indexes.cpp
```diff
@@ -443,6 +444,28 @@ uritoken(AccountID const& issuer, Blob const& uri)
             LedgerNameSpace::URI_TOKEN, issuer, Slice{uri.data(), uri.size()})};
 }
 
+Keylet
+cron(uint32_t timestamp, AccountID const& id)
+{
+    static const uint256 ns = indexHash(LedgerNameSpace::CRON);
```
sublimator, line 471:
```C++
// Constructs an ordered CRON keylet (32 bytes):
//   [8-byte namespace][4-byte timestamp (big-endian, seconds)][20-byte AccountID]
//
// Properties
// - Namespacing: first 8 bytes are the most-significant bytes of indexHash(LedgerNameSpace::CRON).
// - Uniqueness (per ts,acct): exactly ONE cron per (timestamp, AccountID). Insert is upsert (last-write-wins).
//   This is fine because we only ever allow one cron per account at a time—if the same (ts,acct) is written,
//   it’s simply an update to that single entry (idempotent; no duplicate leaves).
// - Iteration order: chronological by timestamp (BE), then by raw AccountID bytes.
//   NOTE: raw AccountID ordering may bias priority; consider hashing AccountID for uniform per-timestamp spread.
// - Expected accidental prefix collisions (foreign objects sharing the 8-byte namespace): n / 2^64,
//   assuming uniform high-64-bit distribution of other objects.
//   Examples: 100M → ~5.4e-12, 1B → ~5.4e-11, 10B → ~5.4e-10, 100B → ~5.4e-9 (negligible).
```

Will have some pointless inner nodes on the way down to this prefix, but I guess that's fine.

src/ripple/protocol/impl/LedgerFormats.cpp
```diff
@@ -366,6 +367,15 @@ LedgerFormats::LedgerFormats()
         },
         commonFields);
 
+    add(jss::Cron,
+        ltCRON,
+        {
+            {sfOwner,                soeREQUIRED},
```
sublimator, line 373: sfOwnerNode ? 

or is that in commonFields?

src/ripple/protocol/impl/Indexes.cpp
sublimator: > //   NOTE: raw AccountID ordering may bias priority; consider hashing AccountID for uniform per-timestamp spread.

Vanity accounts, shrug?
As we know, the less Index hashes the better though :) ?!

sublimator: re: fairness

If only $n many run, what happens to the ones for a given timestamp that aren't executed ? 

At any given time, it always starts at the "cron base" and finds the next key, correct ? 

So jobs are only ever delayed, not canceled

But if there is congestion, it's always going to be the same low accounts that win? They won't get to take turns out of order, but they'll always be first in the order.

Any special hashing that doesn't include the timestamp to randomize won't really help.

src/ripple/app/tx/impl/SetCron.cpp
RichardAH: > TxType ?

It's a typed enum: `protocol/TxFormats.h:56:enum TxType : std::uint16_t`
afaik this is the correct way to do this, the cast does nothing since the enum is already a uint16
```
protocol/impl/STLedgerEntry.cpp:64:        safe_cast<LedgerEntryType>(getFieldU16(sfLedgerEntryType)));
protocol/impl/STTx.cpp:63:    tx_type_ = safe_cast<TxType>(getFieldU16(sfTransactionType));
protocol/impl/STTx.cpp:78:    tx_type_ = safe_cast<TxType>(getFieldU16(sfTransactionType));
protocol/impl/STTx.cpp:94:    tx_type_ = safe_cast<TxType>(getFieldU16(sfTransactionType));
app/misc/impl/TxQ.cpp:1503:                if (safe_cast<TxType>(sle->getFieldU16(sfLedgerEntryType)) ==
app/tx/impl/Transactor.cpp:1439:                safe_cast<TxType>(ctx_.tx.getFieldU16(sfTransactionType)) ==
```

mov-bx-0xb800:
I like the idea, but what's the difference between this and hooks?

src/ripple/protocol/impl/Indexes.cpp
RichardAH: jobs are always cancelled and recreated (later) by the ttCRON doApply. It's possible someone saturates all the crons from 0 to present though, that's an attack vector ... we should consider some maximum age maybe?

sublimator:
I am not 100% sure myself, but it seems this would remove the need to
submit hook triggers via a client. This can automate hooks?

src/ripple/protocol/impl/Indexes.cpp
sublimator: > It's possible someone saturates all the crons from 0 to present though, that's an attack vector

hrm

sublimator: I'm not sure the mechanism for which you could practically saturate from 0 to present without them simply hitting the limits that Fee is supposed to impose, given you can't create crons for in the past. With that said, if this allows people to create transactions/effects/hooks without paying commensurate Fee there could be a problem?

sublimator: > we should consider some maximum age maybe?

Yeah, because you can't simply skip old ones can you, as that's how the delay mechanism works

src/ripple/app/tx/impl/SetCron.cpp
sublimator:
```
if (safe_cast<LedgerEntryType>(sleCron->getFieldU16(sfLedgerEntryType))
```

You probably want the above form to be semantically kosher, given you're later comparing it against it `ltCRON` which is of type `enum LedgerEntryType : std::uint16_t`

src/ripple/protocol/impl/Indexes.cpp
sublimator: I wonder if you could add a rescheduling Fee somehow??

RichardAH:
> I am not 100% sure myself, but it seems this would remove the need to submit hook triggers via a client. This can automate hooks?

It's like setting an alarm so the hook can wake itself up later and do more processing, rather than being poked by an external service.

src/ripple/protocol/impl/LedgerFormats.cpp
RichardAH: Good catch

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    // set operation
+
+    auto delay = ctx.tx.getFieldU32(sfDelaySeconds);
+    if (delay > 31536000UL /* 365 days in seconds */)
+    {
+        JLOG(j.debug())
+            << "SetCron: DelaySeconds was too high. (max 14 days in seconds).";
+        return tecDELAY_OR_REPEAT_COUNT_TOO_LARGE;
+    }
```
tequdev: should be checked in preflight

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (!hasRepeat)
+        return tesSUCCESS;
+
+    auto recur = ctx.tx.getFieldU32(sfRepeatCount);
+    if (recur > 256)
+    {
+        JLOG(j.debug()) << "SetCron: RepeatCount too high. Limit is 256. Issue "
+                           "new SetCron to increase.";
+        return tecDELAY_OR_REPEAT_COUNT_TOO_LARGE;
+    }
```
tequdev: should be checked in preflight

src/ripple/app/tx/impl/SetCron.cpp
RichardAH: true good catch... in an earlier iteration of this code the check needed the current time from the last closed ledger, that's why it's in preclaim... I'll move it now

tequdev (commented):
In the current implementation, I believe the `Fee` for a Cron transaction is set to 0.

When calling a Hook via Cron, a cost should be set. What are your thoughts on this?

RichardAH:
> In the current implementation, I believe the `Fee` for a Cron transaction is set to 0.
> 
> When calling a Hook via Cron, a cost should be set. What are your thoughts on this?

The Cron txn is a pseudo, there's nowhere to claim a fee from (Account 0). However the idea here is the `sfOwner` is only a weak TSH on the cron transaction. That means it's a collect call... the hook account pays the fee.

tequdev:
> The Cron txn is a pseudo, there's nowhere to claim a fee from (Account 0). However the idea here is the `sfOwner` is only a weak TSH on the cron transaction. That means it's a collect call... the hook account pays the fee.

However, if 10,000 accounts that haven't set up Hooks are set to Cron with a very small `DelaySeconds`, wouldn't the network be flooded with Cron transactions?

How about adding `sfRepeatCount * baseFee` in `ttCronSet`'s `calculateBaseFee()`?

RichardAH:
> However, if 10,000 accounts that haven't set up Hooks are set to Cron with a very small `DelaySeconds`, wouldn't the network be flooded with Cron transactions?
> 
> How about adding `sfRepeatCount * baseFee` in `ttCronSet`'s `calculateBaseFee()`?

Good one, the fee to set a cron should be proportional to the repeats indeed. There's a limit of 128 crons per ledger, so even if people flooded it, ledgers will still close

src/ripple/app/tx/impl/SetCron.cpp
RichardAH: gotcha completely missed that it's a ledger entry type not a txtype

src/ripple/protocol/impl/Indexes.cpp
RichardAH: We could randomly order accounts but I feel like we need to do this in a KISS way.

Simplest way to do it is just append the timestamp to the account ID and hash that. A cheaper almost a fair way to do it is take the modulo 32 of the timestamp and rotate the bytes of the Account ID by that number. So if the modulo is 16 then the second half of the account's bytes appear first, followed by the first half.

zgrguric:
If account cron is not among 128 crons in ledger, does that mean accounts cron will be executed next ledger?

src/ripple/protocol/impl/Indexes.cpp
sublimator: Oh, I saw elsewhere the hooks pay the fee as weak TSH 

>  I feel like we need to do this in a KISS way.

Yah

RichardAH:
> If account cron is not among 128 crons in ledger, does that mean accounts cron will be executed next ledger?

yeah the engine just grabs the first 128 crons by searching forward from timestamp=0, and stops when it hits the current time. so every cron that is older than the current time is queued for immediate execution in as many subsequent batches of 128 txns (as many ledgers) as required to clear the backlog

tequdev:
Instead of determining Cron deletion based on the presence of two fields, how about adding a tfDeleteFlags for CronSet?

Using a flag instead of a field combination makes the transaction intent clearer for users and the explorer.

RichardAH:
> Instead of determining Cron deletion based on the presence of two fields, how about adding a tfDeleteFlags for CronSet?
> 
> Using a flag instead of a field combination makes the transaction intent clearer for users and the explorer.

Sure, makes sense. Feel free to change it.

Thanks for the tests btw

src/test/app/Cron_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+        using namespace jtx;
+        using namespace std::literals::chrono_literals;
+
+        // setup env
+        auto const alice = Account("alice");
+        auto const issuer = Account("issuer");
+
+        for (bool const withCron : {false, true})
+        {
+            // If the BalanceRewards amendment is not enabled, you should not be
```
RichardAH: stray comments left over from copypaste

src/test/app/Cron_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+            *this, network::makeNetworkConfig(21337), features | featureCron};
+
+        env.fund(XRP(1000), alice);
+        env.close();
+
+        //----------------------------------------------------------------------
+        // preflight
+
+        // temINVALID_FLAG
+        // can have flag 1 set to opt-out of rewards
```
RichardAH: ^

src/test/app/Cron_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    void
+    testWithFeats(FeatureBitset features)
+    {
+        testEnabled(features);
+        testFee(features);
+        testInvalidPreflight(features);
+        testInvalidPreclaim(features);
+        testDoApply(features);
```
RichardAH, line 443: Thanks for doing the tests but I think we need also to actually test the ttCRON insertion after the delay is satisfied

tequdev: Yes, I've added only the ttCronSet tests for now.
I'm currently working on the ttCron tests.

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,274 @@
...
+{
+    if (ctx.tx.isFlag(tfCronUnset))
+    {
+        auto const account = ctx.tx.getAccountID(sfAccount);
+        auto const sle = ctx.view.read(keylet::account(account));
+        if (!sle)
+            return tefINTERNAL;
+
+        if (!sle->isFieldPresent(sfCron))
+            return tecNO_ENTRY;
```
RichardAH: do we want to tec this? why not tesSUCCESS in the spirit of (if it exists delete it, if it doesn't exist its already deleted)

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+TER
+SetCron::preclaim(PreclaimContext const& ctx)
+{
+    if (ctx.tx.isFlag(tfCronUnset))
+    {
+        auto const account = ctx.tx.getAccountID(sfAccount);
+        auto const sle = ctx.view.read(keylet::account(account));
+        if (!sle)
+            return tefINTERNAL;
```
RichardAH: I was actually wondering if we need `return terNO_ACCOUNT` but it appears its already returned by checkFee earlier in applySteps if it doesn't exist, so this is probably fine. terNO_ACCOUNT is a more consistent error code here though? (and move the check out of the conditional)

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,274 @@
...
+    bool const isDelete = tx.isFlag(tfCronUnset);
+
+    // delay can be zero, in which case the cron will usually execute next
+    // ledger.
+    uint32_t delay{0};
+    uint32_t recur{0};
+
+    if (!isDelete)
+    {
+        delay = tx.getFieldU32(sfDelaySeconds);
```
RichardAH: wouldn't hurt to have a if (!tx.isFieldPresent(sfDelaySeconds)) return tefINTERNAL;`

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,274 @@
...
+        view.update(sle);
+        return tesSUCCESS;
+    }
+
+    // execution to here means we're creating a new Cron object and adding it to
+    // the user's owner dir
+
+    Keylet klCron = keylet::cron(afterTime, id);
+
+    bool const alreadyExists = view.exists(klCron);
```
RichardAH: looking at this logic now, there's no way it could alreadyExist because if it did then sfCron would point at it and we just deleted it? but this logic doesn't hurt other than being slightly confusing.

src/ripple/app/tx/impl/SetCron.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if (!hasRepeat && !hasDelay)
+        // delete cron
+        return baseFee;
+
+    // factor a cost based on the total number of txns expected
+    // for RepeatCount of 0 we have this txn (SetCron) and the
+    // single Cron txn (2). For a RepeatCount of 1 we have this txn,
+    // the first time the cron executes, and the second time (3).
+    uint32_t const additionalExpectedExecutions =
+        tx.getFieldU32(sfRepeatCount) + 1;
```
RichardAH: this is a crash bug @tequdev sfRepeatCount may be absent indicating implicit 0 repeats

src/ripple/app/tx/impl/SetCron.cpp
tequdev: If it finally results in a TEC error, I'd prefer to mark it as a TEC error at the preclaim stage, but this isn't a strong opinion. It's fine to exit with a tesSuccess status.

src/ripple/app/tx/impl/SetCron.cpp
tequdev: rippled often returns `tefINTERNAL` in this case, but in some cases it returns `terNO_ACCOUNT`.


`tefINTERNAL` might be easier to understand as it's an error that should not be reached.

src/ripple/app/tx/impl/SetCron.cpp
tequdev: Although this is unlikely to happen in reality, I will add it as defensive check. (as `tecINTERNAL`)

src/ripple/app/tx/impl/SetCron.cpp
tequdev: `getFieldU32()` returns 0 instead of crashing if the field does not exist, but I will explicitly add an existence check to prevent such questions from arising.

src/ripple/app/tx/impl/SetCron.cpp
RichardAH: I'm just thinking how we can simplify the logic for bots and so on... a tec code means the action failed and needs to be re-tried, whereas tesSUCCESS means the desired outcome was achieved and don't try again. Whether or not the cron existed to be deleted, it now does not exist... which was the objective of the transaction so I think tesSUCCESS works better

tequdev: [c1e011a](https://github.com/Xahau/xahaud/pull/590/commits/c1e011a16a1784d2ac4cb0ba8a75f0705c9afb48)

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1477,6 +1477,64 @@ TxQ::accept(Application& app, OpenView& view)
...
+                }
+            }
+
+            klStart = *next;
+        }
+
+        auto const seq = view.info().seq;
+
+        // insert Cron pseudos for each of the accs we need to ping
+        for (AccountID const& id : cronAccs)
```
RichardAH, line 1518: we need to canonicalize this set before iterating it.

src/ripple/protocol/jss.h
```diff
@@ -50,15 +50,18 @@ JSS(AccountSet);     // transaction type.
 JSS(Amendments);     // ledger type.
 JSS(Amount);         // in: TransactionSign; field.
 JSS(Authorize);      // field
+JSS(Alarm);
```
tequdev:
```suggestion
```

src/ripple/protocol/jss.h
```diff
@@ -95,40 +98,41 @@ JSS(isSigningField);      // out: RPC server_definitions
...
+JSS(RippleState);           // ledger type.
+JSS(SLE_hit_rate);          // out: GetCounts.
+JSS(SetFee);                // transaction type.
+JSS(SetRemarks);            // transaction type
+JSS(UNLModify);             // transaction type.
+JSS(UNLReport);             // transaction type.
+JSS(SettleDelay);           // in: TransactionSign
+JSS(SendMax);               // in: TransactionSign
+JSS(Sequence);              // in/out: TransactionSign; field.
+JSS(SetAlarm);
```
tequdev:
```suggestion
```

and need to fix format (comment indent)

PR #597, opened 2025-10-20
Add hook header generators and CI verification workflow

tequdev:

- Added generate_error.sh, generate_extern.sh generate_tts.sh to generate each header file.
- Organize the order of Hook APIs in applyHook.h.
- Add a GitHub Action to verify that each header file matches what is generated by the script.

### Type of Change

- [x] Chore (no impact to binary, e.g. `.gitignore`, formatting, dropping support for older tooling)


hook/generate_sfcodes.sh
```diff
@@ -1,8 +1,13 @@
-#/bin/bash
-RIPPLED_ROOT="../src/ripple"
+#/bin/sh
```
sublimator: missing `!` shebang

sublimator:
```
       - target: hook/error.h
            generator: ./hook/generate_error.sh
          - target: hook/extern.h
            generator: ./hook/generate_extern.sh
          - target: hook/sfcodes.h
            generator: bash ./hook/generate_sfcodes.sh
          - target: hook/tts.h
            generator: ./hook/generate_tts.sh
            
```

I guess that's why `bash` is required there, though it shouldn't be

sublimator: Should we standardize on using `#!/bin/bash` for all our scripts given it's used for all the other scripts, and we were using it here in one case (`generator: bash ./hook/generate_sfcodes.sh`) ?

hook/generate_error.sh
```diff
@@ -0,0 +1,58 @@
...
+            if (name == "" || value == "")
+                return
+            printf "#define %s %s\n", name, value
+        }
+
+        {
+            line = $0
+            if (line ~ /enum[[:space:]]+hook_return_code/)
+                next
+            if (line ~ /^\s*\{/)
```
sublimator:
```
# On macOS, the bundled awk is BSD awk and requires POSIX-compliant regex.
# GNU awk extension \s doesn't work:
echo "  test" | awk '/^\s+test/ { print "match 1" }'
# (no output)

# POSIX [[:space:]] works:
echo "  test" | awk '/^[[:space:]]+test/ { print "match 2" }'
# match 2
```

hook/generate_sfcodes.sh
tequdev: yeah, good catch

hook/generate_error.sh
sublimator: Do you just install gnu awak as awk ?

tequdev: no, just using the default on macOS and it's works

sublimator: Ok, I was confused, as you responded with a thumbs up

I actually haven't run any of the scripts locally

But AI flagged the "\s" and then the output from above led me to believe it wasn't working on MacOS

tequdev: https://github.com/Xahau/xahaud/pull/597/commits/b7dc8a674932c50404c6ed1cf356ade700866d58

sublimator: Aaaaah, I see

# AWK Bug Analysis: Latent Issues Masked by sed

## The Issue

The AWK patterns ARE triggered, but their bugs don't matter because **sed already filtered the input**.

## What Actually Happens

### Sed sends to AWK:

```cpp
enum hook_return_code : int64_t
  {
    SUCCESS = 0,
    OUT_OF_BOUNDS = -1,
    ...
    TOO_MANY_NAMESPACES = -45
  };
```

**Sed stops here.** The `ExitType` enum never reaches AWK!

### AWK processes those lines:

1. `/enum[[:space:]]+hook_return_code/` → ✅ **Triggers**, skips that line

2. `/^\s*\{/` → ❌ **Fails to match** `  {` on BSD awk (bug!)
   - But doesn't matter: line goes to buffer, `emit()` rejects it (no `=`)

3. `/^\s*\};/` → ❌ **Fails to match** `  };` on BSD awk (bug!)
   - But doesn't matter: no more lines coming from sed anyway!

## Summary

- The patterns **ARE tested**
- The bugs **DO exist**
- The bugs **DON'T matter** because sed already did the filtering

**It's defensive/redundant code with latent bugs that are masked by upstream processing.**

sublimator: tbh, this is why I like Python :)

sublimator (approved):

PR #607, opened 2025-10-24
fix(logs): formatting fixes, color handling, and debug build defaults

sublimator:
Fixes jagged formatting where location info appeared with leading
spaces after blank lines in multiline log messages. Location suffix
now appears immediately after content with trailing newlines preserved.

### 1. Fix jagged location suffix formatting
- Location info `[file:line]` now appears directly after the last non-whitespace character
- Trailing newlines are preserved after the location suffix (not before it)
- Handles whitespace-only messages correctly (skips adding location)

### 2. Optimize hot logging path
The location suffix code runs on every log message when `BEAST_ENHANCED_LOGGING` is enabled:
- Single backward scan using optimized `find_last_not_of()`
- Exact pre-allocation to avoid string reallocations
- Direct string operations instead of ostringstream overhead
- Move semantics for final assignment

### 3. Fix NO_COLOR support for partition names
- Partition colors (e.g., `Application:`, `NetworkOPs:`) were ignoring `NO_COLOR`
- Now both partition names and location suffixes properly respect `should_log_use_colors()`
- Honors `NO_COLOR` environment variable and terminal detection

### 4. Fix BEAST_ENHANCED_LOGGING for Debug builds
- Uses CMake generator expression instead of configure-time check
- Source file wrapped in `#ifdef` guards (compiles to empty when disabled)
- `date-tz` always linked (code is `#ifdef` guarded, linker strips unused)

## Before

image

## After 

image


src/ripple/beast/utility/EnhancedLogging.h
```diff
@@ -41,6 +41,14 @@ get_log_highlight_color();
 constexpr const char*
 strip_source_root(const char* file)
 {
+    // Handle relative paths from build/ directory (common with ccache)
+    // e.g., "../src/ripple/..." -> "ripple/..."
+    if (file[0] == '.' && file[1] == '.' && file[2] == '/' && file[3] == 's' &&
```
RichardAH: it would probably never happen that a string shorter than 7 chars would be passed in here, and it's constexpr but I feel like you should still do a length sanity check?

RichardAH: if (file && file[0]...

sublimator: done

PR #611, opened 2025-10-24
Add Previous fields to `ltCron`

tequdev:
### Type of Change

### API Impact


sublimator (approved):
Good catch

PR #605, opened 2025-10-23
Hook API Refactor1: whitelist api at `Enum.h`

tequdev:
### Type of Change

- [x] Refactor (non-breaking change that only restructures code)

PR #621, opened 2025-10-30
Hook API Refactor2: Amendment Guards

tequdev:
### Context of Change

require: 
- https://github.com/Xahau/xahaud/pull/605

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/hook/Enum.h
```diff
@@ -372,105 +376,129 @@ const double fee_base_multiplier = 1.1f;
...
 
 // RH NOTE: Find descriptions of api functions in ./impl/applyHook.cpp and
 // hookapi.h (include for hooks) this is a map of the api name to its return
 // code (vec[0] and its parameters vec[>0]) as wasm type codes
-static const APIWhitelist import_whitelist{
+inline APIWhitelist
+getImportWhitelist(
+#ifndef GUARD_CHECKER_BUILD
+    Rules const& rules
+#endif
```
RichardAH: I feel like leaving the param in and making it a std::optional is a bit cleaner than changing the definition based on how its compiled?

tequdev: When building guard-checker as an executable binary, it can't reference the `Rules`.

src/ripple/app/hook/Guard.h
```diff
@@ -1020,31 +1021,24 @@ validateGuards(
...
+                bool found_in_whitelist = (it != it_end);
+
                 if (import_name == "_g")
                 {
                     guard_import_number = func_upto;
                 }
-                else if (
-                    hook_api::import_whitelist.find(import_name) ==
-                    hook_api::import_whitelist.end())
+                if (!found_in_whitelist)
```
RichardAH, line 1032: is this logic the same as what it replaces??

tequdev: In `src/test/app/SetHook_test.cpp`, the change confirms that before the Amendment is enabled, `temMALFORMED` is returned, and after it is enabled, the transaction succeeds.

PR #666, opened 2026-01-13
fixXahauGenesisOwnerCount

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/ripple/app/tx/impl/SetHook.cpp
```diff
@@ -1221,6 +1221,23 @@ updateHookParameters(
...
+    if (!hookObj.isFieldPresent(sfHookHash))
+        return 0;
+
+    int reserve{1};
+
+    if (hookObj.isFieldPresent(sfHookParameters))
+        reserve += hookObj.getFieldArray(sfHookParameters).size();
+
+    if (hookObj.isFieldPresent(sfHookGrants))
+        reserve += hookObj.getFieldArray(sfHookGrants).size();
```
RichardAH, line 1242: would be nice to have a way to ensure hookSetObj is never accidentally passed into here, even just via a comment at the top

src/test/app/XahauGenesis_test.cpp
```diff
@@ -2308,7 +2325,9 @@ struct XahauGenesis_test : public beast::unit_test::suite
             BEAST_EXPECT(!!root);
             if (root)
             {
-                BEAST_EXPECT(root->getFieldU32(sfOwnerCount) == mc * 2 + 2);
+                BEAST_EXPECT(
+                    root->getFieldU32(sfOwnerCount) ==
+                    mc * 2 + 2 + paramsCount);
                 BEAST_EXPECT(root->getFieldU32(sfFlags) & lsfDisableMaster);
                 BEAST_EXPECT(root->getAccountID(sfRegularKey) == noAccount());
             }
```
sublimator, line 2333: This assertion currently has a precedence bug and is not checking what it appears to check. As written, `ownerCount == !testFlag ? expectedOwnerCount : 14` is parsed as `(ownerCount == !testFlag) ?
  expectedOwnerCount : 14`, so `BEAST_EXPECT(...)` receives either `expectedOwnerCount` or `14` (both non-zero), which can pass even when `ownerCount` is wrong. The intended check is that owner count
  equals `expectedOwnerCount` when `testFlag == false`, otherwise equals `14`. We need to parenthesize the ternary on the right-hand side: `ownerCount == (!testFlag ? expectedOwnerCount : 14)`.

src/test/app/XahauGenesis_test.cpp
```diff
@@ -247,7 +249,10 @@ struct XahauGenesis_test : public beast::unit_test::suite
         BEAST_EXPECT(
             genesisAccRoot->getFieldAmount(sfBalance) ==
             XahauGenesis::GenesisAmount);
-        BEAST_EXPECT(genesisAccRoot->getFieldU32(sfOwnerCount) == 2);
+        BEAST_EXPECT(
+            genesisAccRoot->getFieldU32(sfOwnerCount) == !testFlag
```
sublimator, line 253: This assertion currently has a precedence bug and is not checking what it appears to check. As written, \ownerCount == !testFlag ? expectedOwnerCount : 14` is parsed as `(ownerCount == !testFlag) ?
  expectedOwnerCount : 14`, so `BEAST_EXPECT(...)` receives either `expectedOwnerCount` or `14` (both non-zero), which can pass even when `ownerCount` is wrong. The intended check is that owner count
  equals `expectedOwnerCount` when `testFlag == false`, otherwise equals `14`. Please parenthesize the ternary on the right-hand side: `ownerCount == (!testFlag ? expectedOwnerCount : 14)`.`

PR #668, opened 2026-01-20
Add tests for Hooks fee

tequdev:

Currently, there's no way to detect if unintended changes are made to the guard-checker that would result in different HookFee calculations.

I'm going to add a test to make this detectable.

Only a test file is modified in this PR.

### Type of Change


RichardAH:
Some rebase issue with this branch so I cherry-picked to a new branch, should be able to merge that one

PR #674, opened 2026-02-02
Build xahau.h from source and check in workflow

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


.github/workflows/check-genesis-hooks.yml
```diff
@@ -0,0 +1,107 @@
...
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add
+          sudo apt-get update
+          sudo apt-get install clang-format-${CLANG_VERSION}
+          clang-format --version
+
+      # Install wasienv (WebAssembly SDK)
+      - name: Install wasienv
+        run: |
+          # Download install.sh
+          curl -o /tmp/wasienv-install.sh https://raw.githubusercontent.com/wasienv/wasienv/master/install.sh
```
RichardAH, line 42: should we pin a version of this maybe?

tequdev: seems like wasienv doesn't use version management, so would it be okay to use the latest commit hash? (The most recent commit was five years ago, though.)


Inside the sh file, the version isn't actually specified since it just uses commands like `pip install wasienv`.

PR #675, opened 2026-02-04
NestedMultiSign

RichardAH:
# NOTE 

Please also see rippled side code @
https://github.com/XRPLF/rippled/pull/6368

And the XLS
https://github.com/XRPLF/XRPL-Standards/discussions/472

----

# NestedMultiSign Amendment

## Overview

This amendment introduces **nested multi-signature validation**, allowing signer lists to delegate signing authority to other accounts that themselves have signer lists. This enables hierarchical organizational structures where, for example, a company account can require signatures from department accounts, which in turn require signatures from authorized individuals.

### Hierarchical Signing (up to 4 levels deep)

- Signers can now delegate to their own signer lists rather than providing a direct signature
- Enables corporate governance structures: `Company → Departments → Teams → Individuals`
- Each level validates against its own signer list and quorum requirements
- Weight is only contributed to the parent level when a signer's own quorum is satisfied

### Cycle Detection & Quorum Relaxation

- Detects circular dependencies in signer list configurations (e.g., A→B→A)
- Prevents permanent fund lockout when accounts accidentally create cyclical signer relationships
- When a cycle makes certain signers unavailable, the effective quorum is relaxed to the maximum achievable weight
- Accounts with `effectiveQuorum == 0` (all signers cyclic) correctly fail with `tefBAD_QUORUM`

## Transaction Structure

Nested signers are represented by including an `sfSigners` array instead of `sfSigningPubKey`/`sfTxnSignature`:

```json
{
  "Signers": [{
    "Signer": {
      "Account": "DepartmentA",
      "Signers": [{
        "Signer": {
          "Account": "Employee1",
          "SigningPubKey": "...",
          "TxnSignature": "..."
        }
      }]
    }
  }]
}
```

## Validation Rules

- Maximum nesting depth: 4 levels
- Signers at each level must be in strict ascending order (consensus requirement)
- A signer entry cannot have both nested signers AND signature fields
- Leaf signers must provide valid `SigningPubKey` and `TxnSignature`
- Each level's accumulated weight must meet its quorum (or relaxed quorum if cycles exist)

## Error Codes

| Code | Condition |
|------|-----------|
| `tefBAD_SIGNATURE` | Depth limit exceeded, invalid signer, or malformed signer entry |
| `tefBAD_QUORUM` | Insufficient weight to meet quorum, or all signers are cyclic |
| `tefNOT_MULTI_SIGNING` | Nested signer account has no signer list configured |
| `tefMASTER_DISABLED` | Leaf signer's master key is disabled |
| `tefINTERNAL` | Invariant violation (should never occur) |

## Backward Compatibility

When `featureNestedMultiSign` is disabled, nested signer arrays are rejected with `temMALFORMED`, preserving existing behavior.


src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -1183,12 +1183,21 @@ transactionSubmitMultiSigned(
...
     if (std::find_if_not(
             signers.begin(), signers.end(), [](STObject const& obj) {
-                return (
-                    // A Signer object always contains these fields and no
-                    // others.
-                    obj.isFieldPresent(sfAccount) &&
-                    obj.isFieldPresent(sfSigningPubKey) &&
-                    obj.isFieldPresent(sfTxnSignature) && obj.getCount() == 3);
+                if (obj.getCount() != 4 || !obj.isFieldPresent(sfAccount))
+                    return false;
```
RichardAH: remove 4 here

RichardAH: if (!obj.isFieldPresent(sfAccount))
                return false;

            // leaf signer
            if (obj.isFieldPresent(sfSigningPubKey) &&
                obj.isFieldPresent(sfTxnSignature) &&
                !obj.isFieldPresent(sfSigners))
                return obj.getCount() == 3;

            // nested signer
            if (!obj.isFieldPresent(sfSigningPubKey) &&
                !obj.isFieldPresent(sfTxnSignature) &&
                obj.isFieldPresent(sfSigners))
                return obj.getCount == 2;

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -369,64 +369,124 @@ STTx::checkMultiSign(
...
+    // Define recursive lambda for checking signatures at any depth
+    std::function<Expected<void, std::string>(
+        STArray const&, AccountID const&, int)>
+        checkSignersArray;
 
-        // No duplicate signers allowed.
-        if (lastAccountID == accountID)
-            return Unexpected("Duplicate Signers not allowed.");
+    checkSignersArray = [&](STArray const& signersArray,
+                            AccountID const& parentAccountID,
```
RichardAH: unused parentAccuontID

zgrguric:
First, thank you for doing this.
What is overhead to xahaud for more than 4 levels? Is there benchmark tests for this?

sublimator:
> What is overhead to xahaud for more than 4 levels? 

Good question, wonder if there should be some kind of leaf specified limit ?

src/libxrpl/protocol/STTx.cpp
```diff
@@ -369,64 +369,124 @@ STTx::checkMultiSign(
...
-        // Verify the signature.
-        bool validSig = false;
-        try
+        for (auto const& signer : signersArray)
         {
-            Serializer s = dataStart;
-            finishMultiSigningData(accountID, s);
+            auto const accountID = signer.getAccountID(sfAccount);
+
+            // The account owner may not multisign for themselves.
```
shortthefomo, line 421: Nested signer matching parent account is not rejected in STTx::checkMultiSign
In the recursive checkSignersArray lambda in STTx::checkMultiSign, the check accountID == txnAccountID only compares against the top-level transaction account ID, not the parentAccountID at each recursion level. The parentAccountID parameter is passed into the lambda but never used.

shortthefomo: But parentAccountID (the account being signed for at the current nesting level) is never checked. This means if account B's signer list includes B itself, and B is a nested signer for account A, the recursive call would pass parentAccountID = B but never reject B appearing in its own nested signer array.
In the original non-recursive code, this wasn't an issue because there was only one level. Now with nesting, a signer at depth N could list themselves as a sub-signer at depth N+1, which should be rejected but isn't.
Note: Transactor::checkMultiSign has cycle detection via the ancestors set which would catch this at the preclaim stage, but STTx::checkMultiSign (the preflight signature verification) does not, allowing malformed transactions to pass signature verification and consume more processing resources than necessary.

src/libxrpl/protocol/InnerObjectFormats.cpp
```diff
@@ -44,8 +44,9 @@ InnerObjectFormats::InnerObjectFormats()
         sfSigner.getCode(),
         {
             {sfAccount, soeREQUIRED},
-            {sfSigningPubKey, soeREQUIRED},
-            {sfTxnSignature, soeREQUIRED},
+            {sfSigningPubKey, soeOPTIONAL},
```
shortthefomo, line 52: Making sfSigningPubKey and sfTxnSignature optional changes validation for ALL signers, not just nested
The change in InnerObjectFormats.cpp:47-49 from soeREQUIRED to soeOPTIONAL for sfSigningPubKey and sfTxnSignature in the sfSigner inner object format is a global change that affects all signer deserialization, not just when featureNestedMultiSign is enabled. This means that even without the amendment enabled, a malformed signer object missing sfSigningPubKey or sfTxnSignature will now successfully deserialize (whereas before it would throw during applyTemplate). The runtime checks in STTx::checkMultiSign (STTx.cpp:445-449) and Transactor::checkMultiSign (Transactor.cpp:1101-1107) do validate that leaf signers have these fields, so this is caught later. However, the test in STTx_test.cpp:1797-1803 (Test case 2) had to be commented out because it tested that deserialization itself would reject a signer without sfSigningPubKey. This is a defense-in-depth regression—invalid data that was previously rejected at the serialization layer now passes through to application-level validation.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -289,8 +289,40 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
...
     // for the transaction.
-    std::size_t const signerCount =
-        tx.isFieldPresent(sfSigners) ? tx.getFieldArray(sfSigners).size() : 0;
+    std::size_t signerCount = 0;
+    if (tx.isFieldPresent(sfSigners))
+    {
+        // Define recursive lambda to count all leaf signers
+        std::function<std::size_t(STArray const&)> countSigners;
+
+        countSigners = [&](STArray const& signers) -> std::size_t {
```
shortthefomo: Fee calculation counts only leaf signers, not nesting overhead
The recursive countSigners lambda in Transactor.cpp:298-322 counts only leaf signers (those with sfSigningPubKey and sfTxnSignature). This means the fee for a nested multi-signed transaction is the same as a flat multi-signed transaction with the same number of leaf signers. However, nested transactions require significantly more validation work (recursive signer list lookups, cycle detection, multiple quorum checks). This could be an underpricing concern where deeply nested transactions consume disproportionate validation resources relative to their fee.

src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -1183,12 +1183,21 @@ transactionSubmitMultiSigned(
     // The Signers array may only contain Signer objects.
     if (std::find_if_not(
             signers.begin(), signers.end(), [](STObject const& obj) {
-                return (
-                    // A Signer object always contains these fields and no
-                    // others.
-                    obj.isFieldPresent(sfAccount) &&
-                    obj.isFieldPresent(sfSigningPubKey) &&
-                    obj.isFieldPresent(sfTxnSignature) && obj.getCount() == 3);
+                if (obj.getCount() != 4 || !obj.isFieldPresent(sfAccount))
```
shortthefomo: getCount() != 4 check in TransactionSign.cpp is correct but non-obvious
The check obj.getCount() != 4 at TransactionSign.cpp:1186 initially appears wrong since a leaf signer has 3 explicit fields and a nested signer has 2. However, this is correct because STObject::applyTemplate (in STObject.cpp:116-153) adds nonPresentObject entries for all optional fields not present in the source data. Since the sfSigner template now has 4 fields (sfAccount required, sfSigningPubKey optional, sfTxnSignature optional, sfSigners optional), getCount() always returns 4 regardless of which optional fields are actually set. The old code used getCount() == 3 because the old template had exactly 3 fields (all required). This is fragile coupling between the validation logic and the template definition—if the template changes again, this hardcoded 4 would silently break.

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -923,157 +955,282 @@ NotTEC
...
+        {
+            JLOG(ctx.j.trace())
+                << "checkMultiSign: Cyclic signer detected: " << acc;
+            return tesSUCCESS;
+        }
 
-        // Attempt to match the SignerEntry with a Signer;
-        while (iter->account < txSignerAcctID)
+        // Check depth limit
+        if (depth > maxDepth)
```
shortthefomo, line 1031: No recursion depth limit on STObject deserialization for nested sfSigners
The sfSigner inner object format now includes sfSigners as an optional field (InnerObjectFormats.cpp:49), which means sfSigner objects can recursively contain sfSigners arrays. STObject::set at STObject.cpp:180 has a depth limit of 10, so deeply nested structures won't cause stack overflow during deserialization. However, the STTx checkMultiSign limits depth to 4, while the serializer allows up to 10 levels of nesting. This means an attacker could craft a transaction with 10 levels of nested signers that would successfully deserialize and consume resources during signature checking before being rejected at depth 5. The computational cost grows exponentially with nesting depth since each level can have up to 32 signers.

sublimator: @shortthefomo 

Please see the latest code @
https://github.com/XRPLF/rippled/pull/6368

And the XLS
https://github.com/XRPLF/XRPL-Standards/discussions/472

This PR is stale

Regards

src/libxrpl/protocol/InnerObjectFormats.cpp
sublimator: TODO

PR #677, opened 2026-02-10
fix: subscription webhook delivery stalling on HTTP errors

sublimator:
---

<sub>Last updated: 2026-06-15 | branch: subscription-hooks-fix | commit: 15181211d (Fix truncated RPC response handling)</sub>

---

## Summary

Fix subscription webhook (`url`) delivery permanently stalling when endpoints
return HTTP errors (500, 404, etc.) or responses without `Content-Length`,
affecting **all** subscribers.

**Upstream issue:** [XRPLF/rippled#6341](https://github.com/XRPLF/rippled/issues/6341)
**Upstream PR:** [XRPLF/rippled#6344](https://github.com/XRPLF/rippled/pull/6344)
(this branch ports that work and adds further hardening — flagged
**[beyond #6344]** below, worth folding back upstream).

> Paths moved from `src/ripple/net/impl/` to `src/xrpld/net/detail/` when
> `dev` was merged in; snippets below are extracted from the current tree.

### Bug 1: RPCSub unbounded concurrency

`sendThread()` passed the app's shared `io_service` to `fromNetwork()`, which
posts async handlers and **returns immediately** — no `.run()`, no waiting:

📍 [`src/xrpld/net/detail/RPCCall.cpp:1901-1924`](https://github.com/Xahau/xahaud/blob/12e8e79b1f305f5c7645844e6c8732489e8c82bc/src/xrpld/net/detail/RPCCall.cpp#L1901-L1924)
```cpp
1901     HTTPClient::request(
1902         bSSL,
1903         io_service,
1904         strIp,
1905         iPort,
1906         std::bind(
1907             &RPCCallImp::onRequest,
1908             strMethod,
1909             jvParams,
1910             headers,
1911             strPath,
1912             std::placeholders::_1,
1913             std::placeholders::_2,
1914             j),
1915         RPC_REPLY_MAX_BYTES,
1916         RPC_NOTIFY,
1917         std::bind(
1918             &RPCCallImp::onResponse,
1919             callbackFuncP,
1920             std::placeholders::_1,
1921             std::placeholders::_2,
1922             std::placeholders::_3,
1923             j),
1924         j);
```

The entire deque was drained at full speed, firing one async HTTP connection
per event. Under sustained errors, each connection held an FD for 30s. At
100+ events/ledger every ~4s, this quickly exhausted the 1024 FD budget.

**Fix:** a **local `io_service`** per batch (same pattern as `rpcClient()`)
with bounded concurrency (32 in-flight) and a configurable queue cap
(16384 events by default), dropping on overflow while advancing `seq` so
consumers can detect the gap:

📍 [`src/xrpld/net/detail/RPCSub.cpp:77-115`](https://github.com/Xahau/xahaud/blob/12e8e79b1f305f5c7645844e6c8732489e8c82bc/src/xrpld/net/detail/RPCSub.cpp#L77-L115)
```cpp
  77 void
  78     send(Json::Value const& jvObj, bool broadcast) override
  79     {
  80         std::lock_guard sl(mLock);
  81 
  82         if (mDeque.size() >= maxQueueSize_)
  83         {
  84             // Always advance mSeq so consumers can detect the gap, but
  85             // rate-limit the log: a hopelessly behind endpoint drops on
  86             // every send() and would otherwise flood the log. Warn on
  87             // the first drop of a run and then once per dropLogInterval.
  88             if (mDropped++ % dropLogInterval == 0)
  89             {
  90                 JLOG(j_.warn())
  91                     << "RPCCall::fromNetwork drop: queue full ("
  92                     << mDeque.size() << "), seq=" << mSeq
  93                     << ", endpoint=" << mIp << ", dropped=" << mDropped;
  94             }
  95             ++mSeq;
  96             return;
  97         }
  98 
  99         // Endpoint caught up enough to accept again; reset so the next
 100         // overflow burst logs its first drop immediately.
 101         mDropped = 0;
 102 
 103         auto jm = broadcast ? j_.debug() : j_.info();
 104         JLOG(jm) << "RPCCall::fromNetwork push: " << jvObj;
 105 
 106         mDeque.push_back(std::make_pair(mSeq++, jvObj));
 107 
 108         if (!mSending)
 109         {
 110             // Start a sending thread.
 111             JLOG(j_.info()) << "RPCCall::fromNetwork start";
 112 
 113             startSendingJob();
 114         }
 115     }
```

`sendThread()` processes **one batch per job and re-queues** if more remain,
so a slow endpoint can't hold a worker for the whole backlog (see *worker
starvation* below), and resets `mSending` under the lock on every exit path
so the queue can never wedge (the #6341 failure mode).

### Bug 2: HTTPClient EOF completion leak

When an HTTP response has no `Content-Length` header, `HTTPClient` reads until
EOF. The EOF path in `handleData` logged "Complete." but **never called
`invokeComplete()`** — so the completion callback never fired, the deadline
timer was never cancelled, and (on the shared `io_service`) the
`HTTPClientImp` was kept alive by its `shared_from_this()` captures. Many
frameworks omit `Content-Length` on error responses, so failing webhook
endpoints hit this constantly.

📍 [`src/xrpld/net/detail/HTTPClient.cpp:447-478`](https://github.com/Xahau/xahaud/blob/c1488a7f0/src/xrpld/net/detail/HTTPClient.cpp#L447-L478)
```cpp
 447     void
 448     handleData(
 449         const boost::system::error_code& ecResult,
 450         std::size_t bytes_transferred)
 451     {
 452         if (!mShutdown)
 453             mShutdown = ecResult;
 454 
 455         if (mShutdown && mShutdown != boost::asio::error::eof)
 456         {
 457             JLOG(j_.trace()) << "Read error: " << mShutdown.message();
 458 
 459             invokeComplete(mShutdown);
 460             return;
 461         }
 462 
 463         // Either the read completed normally or it ended at EOF. EOF is a
 464         // successful completion for EOF-delimited responses, but it is an
 465         // error when the server promised a Content-Length and closed early.
 466         JLOG(j_.trace()) << "Complete.";
 467 
 468         mResponse.commit(bytes_transferred);
 469         std::string strBody{
 470             {std::istreambuf_iterator<char>(&mResponse)},
 471             std::istreambuf_iterator<char>()};
 472 
 473         auto completeEc = ecResult;
 474         if (completeEc == boost::asio::error::eof && !mReceivedContentLength)
 475             completeEc.clear();
 476 
 477         invokeComplete(completeEc, mStatus, mBody + strBody);
 478     }
```

The EOF path now delivers the body and invokes completion. EOF is translated to
success only for EOF-delimited responses with no `Content-Length`; if a server
promises `Content-Length` and closes early, EOF remains a transport error.
`RPCCall::onResponse` now rejects nonzero transport errors before parsing a
partial body as JSON, so CLI RPC cannot accept a truncated-but-parseable reply.
This also changes `HTTPClient::get()` behavior for no-`Content-Length`
responses: a clean EOF-delimited body is now a success rather than a fallback
error.

### Use-after-free on concurrent unsubscribe

The sending job captured a raw `this`. An admin `Unsubscribe` runs
`NetworkOPsImp::tryRemoveRpcSub`, which erases the sub from `mRpcSubMap` under
`mSubLock` and destroys it **inline** — while `sendThread` may be running on a
job-queue worker under the unrelated `InfoSub::mLock`. The blocking
`io_service.run()` (up to the 30s request deadline for a hung endpoint) makes
the window large, and a hung endpoint is exactly this fix's scenario.
`RPCSubImp` now derives `enable_shared_from_this` and the job captures
`weak_from_this()`, re-locking it on entry so the sub stays alive for the
batch; once it's gone the job is a no-op:

📍 [`src/xrpld/net/detail/RPCSub.cpp:151-160`](https://github.com/Xahau/xahaud/blob/12e8e79b1f305f5c7645844e6c8732489e8c82bc/src/xrpld/net/detail/RPCSub.cpp#L151-L160)
```cpp
 151 void
 152     startSendingJob()
 153     {
 154         std::weak_ptr<RPCSubImp> weak = weak_from_this();
 155         mSending = m_jobQueue.addJob(
 156             jtCLIENT_SUBSCRIBE, "RPCSub::sendThread", [weak]() {
 157                 if (auto self = weak.lock())
 158                     self->sendThread();
 159             });
 160     }
```

### Job-queue worker starvation

Draining the entire backlog (up to 16384 events, 512 batches × up to 30s) in
one job blocked a worker thread the whole time; with a small pool shared with
consensus/ledger/RPC, a few hung subscribers could occupy most workers.
`sendThread` now processes **one batch per job and re-queues** if more remain,
releasing the worker between batches.

### `HTTPClient::get()` self-reference leak

`get()` bound `shared_from_this()` into `mBuild` (a member), forming a
`this → mBuild → shared_ptr<this>` cycle that never broke — leaking the object
and its socket FD after the request completed. `mBuild` is only invoked from
`handleRequest()`, which always runs inside a `shared_from_this()`-holding
async handler, so a non-owning `this` is safe (pre-existing; also present in
rippled).

### Bounded webhook response buffer

`fromNetwork` hardcoded a 256MB read budget for responses with no
`Content-Length`; each in-flight EOF-path delivery pre-allocated a ~256MB
streambuf (up to 32 × 256MB = 8GB transient). Webhook deliveries (method
`"event"`) ignore the response body, so they are now capped at 1MB; genuine
RPC replies (CLI) keep the 256MB budget.

### Truncated CLI RPC replies rejected

`RPCCallImp::onResponse` used to parse `strData` even when `HTTPClient`
reported a nonzero transport error. A server that declared `Content-Length`,
sent a shorter but valid JSON body, and closed early could therefore be
reported as a successful CLI RPC response. `HTTPClient` now preserves EOF as an
error for truncated `Content-Length` replies, and `onResponse` rejects that
error before parsing.

## Tests & coverage

- **`HTTPClient_test`** (17 cases): resource cleanup across success, HTTP 500,
  connection refused, timeout, server-close, concurrent requests,
  EOF-without-Content-Length, read-error-mid-body, persistent `io_service`,
  and the `get()` self-reference cycle. It also covers EOF-without-
  `Content-Length` success, truncated `Content-Length` error propagation, and
  both the `RPCCall::fromNetwork` happy path and truncated-reply rejection. The
  mock waits for the *client* to close (EOF) before decrementing, so it
  verifies client-side FD release, not just its own lifecycle.
- **`RPCSub_test`** (4 cases): delivery, delivery continues despite HTTP 500
  responses (EOF-delimited — the #6341 stall, end to end), sending restarts
  after the queue drains, and queue-cap drops. Teardown waits on
  `JobQueue::getJobCountTotal(jtCLIENT_SUBSCRIBE) == 0` so the sub isn't
  destroyed while `sendThread` runs.
- Diff/patch coverage clears the codecov gate (gcc-13).

## Test Plan

- [x] Build succeeds (Release + Debug/coverage)
- [x] `ripple.net.HTTPClient` passes (17 cases, 0 failures)
- [x] `ripple.net.RPCSub` passes (4 cases, 0 failures)
- [x] Connected to Xahau mainnet with `x-testnet create-config --network mainnet --hooks-server`
- [x] Ran `x-testnet hooks-server --error 500:0.5` (50% HTTP 500 responses)
- [x] Confirmed events continue flowing through errors
- [ ] Review: consider adding backoff/limit for repeatedly failing endpoints (follow-up)


src/xrpld/net/detail/RPCSub.cpp
```diff
@@ -200,8 +258,15 @@ class RPCSubImp : public RPCSub
 
     int mSeq;  // Next id to allocate.
 
+    std::size_t mDropped = 0;  // Consecutive drops while queue is full.
```
RichardAH, line 261: probably should be uint64_t

PR #698, opened 2026-03-03
Add test for rejected Remit to AMM account

tequdev:

In Remit.cpp, the current code already rejects it based on the existence of lsfDepositAuth, but I've added code to return `tecNO_PERMISSION` when an AMMID exists at the destination to make this more explicit.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/xrpld/app/tx/detail/Remit.cpp
```diff
@@ -288,6 +288,11 @@ Remit::doApply()
         (flags & lsfDisallowIncomingRemit))
         return tecNO_PERMISSION;
 
+    // AMMs can never receive an XRP payment.
```
RichardAH: should be xah... but also how can someone top up the account root's xah if this is forbidden? the accroot might have hooks and crons on it that need gas money

tequdev: I don't think we should allow fees to be paid from AMM accounts, as it would cause the AMM pool to become unbalanced.

The XAH balance of an AMM pool refers directly to the balance in the AccountRoot and isn't managed by any other objects.

RichardAH: true and there's no pathway to "own" an amm account anyway, they're created by the pair so I think this is fine

PR #701, opened 2026-03-04
fix: resolve switch fall-through in util_keylet unimplemented cases

### LINKED ISSUE #700: Case statement fall-through and warnings
Source: https://github.com/Xahau/xahaud/blob/8cfee6c8a318a2476199374aa7f46aef0b4e50ee/src/xrpld/app/hook/detail/applyHook.cpp#L2733

This code now generates (VALID) warnings about implicit fall-through. All the code from that point forward is useless: it will keep falling through case after case, unitl some amendment isn't enabled, causing it to return `INVALID_ARGUMENT`. And if no such amendment is found, it will fall out of the end and return `INVALID_ARGUMENT`.

sublimator:
## Summary
- Fix implicit fall-through in `util_keylet` switch statement for unimplemented keylet types
- Group unimplemented cases with explicit `return INVALID_ARGUMENT`

Closes #700


shortthefomo:
TICKET keylet_code enum value has no switch case (pre-existing)
The keylet_code enum in Enum.h:291 defines TICKET = 13, but there is no corresponding case keylet_code::TICKET: in the switch statement in util_keylet. Any call with keylet type TICKET would fall through to the default path and hit return INVALID_ARGUMENT at applyHook.cpp:2759, which is likely the intended behavior for an unimplemented type. 

However, unlike the other unimplemented types (BRIDGE, MPTOKEN, etc.) which are now explicitly listed with a comment explaining they're not yet implemented, TICKET is silently unhandled. This is pre-existing and outside the scope of this PR, but may be worth adding to the explicit list for consistency and documentation purposes.

tequdev:
Please keep in mind that we need an Amendment guard if we adding implementations as shown in the previous code.

sublimator:
> Please keep in mind that we need an Amendment guard if we adding implementations as shown in the previous code.


You're referring to full bodied implementations, right? 

I simplified the logic to directly return INVALID_ARGUMENT for the non supported amendments. The cases are grouped and separated by a comment with the amendment name for later reference. 

This works around the reported warnings which distracted people, even if it's arguably not strictly needed.

PR #704, opened 2026-03-05
Add new TSH tests

tequdev:

We'll use transactions.macro to ensure that TSH tests are implemented for all transaction types.

If a new transaction type is added without a corresponding test, it will now trigger a build error.

### Type of Change


tequdev:
In some test cases, I added the code like `BEAST_EXPECT(!features[featureDID]);`. This is meant to trigger a test failure when the feature is set to "Supported: yes," indicating that we'll need to update the TSH or add more tests at that point.

Also, by using transaction.macro to define TSH test cases, I have made it so that we can notice if we forget to add TSH-related test code when adding a new transaction type.

PR #708, opened 2026-03-09
Add `--definitions` CLI flag to output static server definitions without starting server

tequdev:
### Type of Change

### API Impact


.github/workflows/xahau-ga-nix.yml
```diff
@@ -375,3 +375,15 @@ jobs:
...
+        if: matrix.compiler_id == 'gcc-11-libstdcxx'
+        run: |
+          ${{ env.build_dir }}/rippled --definitions | python3 -m json.tool > server_definitions.json
+
+      - name: Upload server definitions
+        if: matrix.compiler_id == 'gcc-11-libstdcxx'
+        uses: actions/upload-artifact@v4
+        with:
+          name: server-definitions
+          path: server_definitions.json
```
sublimator, line 478: oooooh!

sublimator (dismissed):

sublimator:
Some half baked musings: 

Should we just add a new executable target ? 
That has a server-definitions command and encode/decode commands ? 


```
Usage: xrpl-codec <command> [options]

Commands:
  server-definitions        Print server definitions JSON
  encode <json>             Encode JSON to hex
  decode <hex>              Decode hex to JSON

Options:
  --codec-type <type>       Serialization type (default: stobject)
                            Supported: stobject, amount, currency, issue, number

Input can be provided as an argument or via stdin.

Examples:
  xrpl-codec server-definitions
  xrpl-codec encode '{"Account": "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"}'
  xrpl-codec decode 811439249EE0886DE835D4F4D47DA9D9B1D2AED83C11
  xrpl-codec encode --codec-type amount '"1000000"'
  echo '{"Fee": "10"}' | xrpl-codec encode
```

Something along these lines ( I had Claude create a little throwaway branch util for updating some binary codec code in various libs ) 

But it might actually be useful ? 

Or maybe just a `codec` subcommand to xahaud?

```
codec encode 
codec decode
codec server-definitions 
```

tequdev:
> Some half baked musings:
> 
> Should we just add a new executable target ? That has a server-definitions command and encode/decode commands ?
> 
> ```
> Usage: xrpl-codec <command> [options]
> 
> Commands:
>   server-definitions        Print server definitions JSON
>   encode <json>             Encode JSON to hex
>   decode <hex>              Decode hex to JSON
> 
> Options:
>   --codec-type <type>       Serialization type (default: stobject)
>                             Supported: stobject, amount, currency, issue, number
> 
> Input can be provided as an argument or via stdin.
> 
> Examples:
>   xrpl-codec server-definitions
>   xrpl-codec encode '{"Account": "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"}'
>   xrpl-codec decode 811439249EE0886DE835D4F4D47DA9D9B1D2AED83C11
>   xrpl-codec encode --codec-type amount '"1000000"'
>   echo '{"Fee": "10"}' | xrpl-codec encode
> ```
> 
> Something along these lines ( I had Claude create a little throwaway branch util for updating some binary codec code in various libs )
> 
> But it might actually be useful ?
> 
> Or maybe just a `codec` subcommand to xahaud?
> 
> ```
> codec encode 
> codec decode
> codec server-definitions 
> ```

I don't think we'll ever need to build xahaud locally just to encode/decode, so I think it would be better to create a CLI in js.

sublimator:
> I don't think we'll ever need to build xahaud locally just to encode/decode, so I think it would be better to create a CLI in js.


Yeah, probably not, but the point of it is maintaining other libs such as js with updated types from the canonical impl

You can for example hand it off to an AI agent to do most of the work

But it's also trivial to create if one needs, so no need to commit it

PR #710, opened 2026-03-11
Fix LedgerNameSpace to ensure uniqueness

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/libxrpl/protocol/Indexes.cpp
```diff
@@ -80,14 +82,14 @@ enum class LedgerNameSpace : std::uint16_t {
     UNL_REPORT = 'R',
     CRON = 'L',
     AMM = 'A',
-    BRIDGE = 'H',
+    BRIDGE = LEDGER_NAMESPACE2(0x01, 'H'),
```
a-noni-mousse, line 85: This is a breaking change if objects of this type exist on the Xahau ledger. Same for CREDENTIAL, DID and ORACLE.

a-noni-mousse (commented):
I don't like this very much. In addition to my comment about potential breaking, I think the weird #define just complicated and confuses thing. And while we're at it, why even a macro? If you feel compelled to have something, why not:

```
constexpr std::uint16_t make_lns(char hi, char low)
{
    return static_cast<std::uint16_t>(lo) + (256 * static_cast<std::uint16_t>(hi));
}
```

But again, why not just do `BRIDGE = 0x1978` or whatever?

src/libxrpl/protocol/Indexes.cpp
tequdev: I'm not particularly set on the macro implementation, so I'm happy to change that.

Regarding the potential breaking changes, this won't affect the Amendment as long as it's either merged before the next release or released with "supported:no" for now.

RichardAH:
There's probably a cleaner solution but this is fine for now

PR #713, opened 2026-03-13
chore: replace levelization shell script with python

sublimator:
Backport of https://github.com/XRPLF/rippled/pull/6325


tequdev (approved):

tequdev:
It used to take about 50 seconds, but it's now down to less than a second.

PR #718, opened 2026-03-16
Named Hook

tequdev:

Add support for named hooks via a new optional HookName field on hook definitions/usages.

This change introduces the featureNamedHooks amendment, allows SetHook to create/install/update hooks with an optional UTF-8 hook name, and updates hook execution/fee calculation so named hooks only run when the incoming transaction includes the matching HookName. The PR also adds coverage for invalid names, create/install/update flows, strong/weak execution paths, and callback/emitted transaction behavior.

### Context of Change

Today hooks are selected only by install position and existing hook matching rules. This change adds a lightweight name-based gate so an installed hook can require an explicit transaction opt-in through HookName.

The implementation keeps HookName on the installed ltHook object rather than the shared hook definition, which avoids changing hook bytecode identity and lets different accounts install the same hook definition with or without a name. Validation enforces that HookName is either empty or a valid UTF-8 string between 4 and 16 bytes (8 to 32 hex chars in JSON form). Transactions carrying HookName are preflight-validated.

### Type of Change

## Before / After

Before:

- SetHook did not accept `HookName`.
- Transactions could not declare a hook name target.

After:

- SetHook accepts optional HookName when featureNamedHooks is enabled.
- Transactions may include optional HookName.
- Installed hooks with a non-empty HookName execute only when the transaction carries the same name.
- Hook chain fee calculation now follows the same name-matching rule.
- Empty HookName on update removes the stored name from the installed hook.


src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -609,6 +621,23 @@ SetHook::validateHookSetEntry(SetHookCtx& ctx, STObject const& hookSetObj)
     }
 }
 
+bool
+SetHook::validateHookName(Blob const& name, beast::Journal const& j)
+{
+    if (name.size() != 0 && (name.size() < 4 || 16 < name.size()))
+    {
+        JLOG(j.trace())
+            << "sfHookName must be between 4 and 16 hex characters.";
```
RichardAH: wouldn't this be 8 and 32 hex characters?

tequdev: true, fixed

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -266,8 +277,24 @@ Transactor::calculateHookChainFee(
...
         if (!hookDef)
         {
+            // LCOV_EXCL_START
             printf("calculateHookChainFee edge case\n");
             continue;
+            // LCOV_EXCL_STOP
+        }
+
+        std::optional<Blob> requiredHookName;
+        if (hookObj.isFieldPresent(sfHookName) &&
```
RichardAH, line 287: is this right? do we need to check the hook def as well?

tequdev: The absence of `sfHookName` in `ltHookDefinition` is intentional.

This is to prevent falling back in cases where `sfHookName` exists in `ltHookDefinition` but the user does not want to use that hook as a Named Hook.

As an alternative, it is possible to set a zero-length Blob in `ltHook[n].sfHookName`. However, I believe it is better to have a design where users do not need to be conscious of HookName settings normally and only use HookName when necessary.

PR #729, opened 2026-04-14
Fix HookAPI  Expected and Refactor Enum classes

tequdev:

Key changes
- [fix bad expected returns and refacor HookAPI::parse_error to enum class](https://github.com/Xahau/xahaud/pull/729/changes/ba4e9dd909f01d340c6fe38ebeeddb75b820a81d)
- [std::variant<UNSIGNED_TYPE(R), hook_api::hook_return_code>](https://github.com/Xahau/xahaud/pull/729/changes/2536f0b136d845c148cefd3f62c63cfdabde4faa)

### Type of Change


src/xrpld/app/hook/detail/applyHook.cpp
```diff
@@ -3560,30 +3462,35 @@ DEFINE_HOOK_FUNCTION(
...
                   << ": Float 0*10^(0) <ZERO>";
-        return 0;
+        return 0ULL;
     }
 
-    uint64_t man = get_mantissa(float1);
-    int32_t exp = get_exponent(float1);
+    auto const man = get_mantissa(float1);
+    if (!man)
+        return 0ULL;
```
sublimator: This changes trace behavior for invalid negative XFL inputs. Previously `get_mantissa(-1)` returned the `INVALID_FLOAT` sentinel, which then failed the range check below and emitted `Float <INVALID>`. Now `get_mantissa`/`get_exponent` return `Unexpected(INVALID_FLOAT)`, and these early `return 0ULL` paths skip that diagnostic. Can we route `!man || !exp` into the existing invalid-log branch instead of returning before it?

(codex review)

src/xrpld/app/hook/detail/applyHook.cpp
```diff
@@ -598,159 +598,46 @@ getTransactionalStakeHolders(STTx const& tx, ReadView const& rv)
...
-             // it is not then we will return the final slot every time.
-             hookCtx.slot_counter <= hook_api::max_slots);
-
-    if (hookCtx.slot_counter > hook_api::max_slots)
-        return {};
-
-    return slot_into;
-}
+// inline int32_t
+// no_free_slots(hook::HookContext& hookCtx)
```
sublimator: We probably don't want this any more?

src/xrpld/app/hook/detail/applyHook.cpp
tequdev: fixed

src/xrpld/app/hook/detail/applyHook.cpp
tequdev: deleted

sublimator (approved):

PR #744, opened 2026-05-11
fix: Ensures canonical order for `PriceDataSeries` upon `PriceOracle` creation  (#5485)

tequdev:
This change fixes an issue where the order of `PriceDataSeries` was out of sync between when `PriceOracle` was created and when it was updated. Although they are registered in the canonical order when updated, they are created using the order specified in the transaction; this change ensures that they are also registered in the canonical order when created.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


RichardAH:
good to get this in before anyone has a chance to vote in oracle amendment

PR #750, opened 2026-05-20
Generate hook extern declarations with C linkage

tequdev:

When developing hooks in languages other than C, such as with cpphooks, referencing the repository's header files will result in build errors if C linkage is missing. Therefore, I will modify the generation script to add C linkage.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/hook/extern.h
+++ b/hook/extern.h
@@ -2,31 +2,34 @@
 // Generated using generate_extern.sh
 #include <stdint.h>
 #ifndef HOOK_EXTERN
+#ifdef __cplusplus
+extern "C" {
+#endif
 
-extern "C" int32_t __attribute__((noduplicate))
+extern int32_t __attribute__((noduplicate))
 _g(uint32_t guard_id, uint32_t maxiter);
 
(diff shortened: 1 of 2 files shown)


RichardAH:
`extern "C"` breaks C compilers? Either generate two headers (hpp and h) or use a macro wrapper like
```
#ifdef __cplusplus
#define EXTERN extern "C"
#else
#define EXTERN extern
#endif
```

tequdev:
> `extern "C"` breaks C compilers? Either generate two headers (hpp and h) or use a macro wrapper like
> 
> ```
> #ifdef __cplusplus
> #define EXTERN extern "C"
> #else
> #define EXTERN extern
> #endif
> ```

Ah, true.

how about this format?

```h
#ifdef __cplusplus
extern "C" {
#endif

...
extern int64_t
accept(uint32_t read_ptr, uint32_t read_len, int64_t error_code);
...

#ifdef __cplusplus
}
#endif
```

It should result in fewer changes.

PR #756, opened 2026-06-12
Fixhookmap

RichardAH:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/test/app/HookAPI_test.cpp
```diff
@@ -3773,6 +3774,60 @@ class HookAPI_test : public beast::unit_test::suite
...
+            // setup a hook on alice, and on claire, no grants
+            env(hook(alice, {{hso(genesis::AcceptHook)}}, 0), fee(XRP(1)));
+            env(hook(claire, {{hso(genesis::AcceptHook)}}, 0), fee(XRP(1)));
+            env.close();
+
+            // First modification
+            auto result1 =
+                api.state_foreign_set(testKey, testNs, aliceid, testData);
+            BEAST_EXPECT(result1.has_value());
+
```
tequdev, line 3806: Just to be safe, might be better toverify that those caches (`hookCtx.result.stateMap`) are created in `hookCtx` and are being carried over to `hookCtx2`.

tequdev (dismissed):
I checked it a few days ago, and since nothing seems to have changed since then, I think it should be fine.

PR #757, opened 2026-06-12
Fix ammLPHolds logic to include escrowed cases

tequdev:
### Type of Change


src/xrpld/app/misc/detail/AMMUtils.cpp
```diff
@@ -148,6 +148,31 @@ ammLPHolds(
...
+                // LCOV_EXCL_START
+                JLOG(j.error())
+                    << "SpendableBalance has illegal value in accountHolds "
+                    << spendableBalance;
+                amount.clear(Issue{currency, ammAccount});
+                // LCOV_EXCL_STOP
+            }
+            else
+                amount = spendableBalance;
+        }
```
RichardAH, line 174: 👍

PR #760, opened 2026-06-17
[PWABoot] SetBoot — on-chain PWA boot-blob (draft PoC, discussion #759)

Hugegreencandle:
**Draft / PoC for design discussion #759** — https://github.com/Xahau/xahaud/discussions/759

A storage primitive for a small on-chain **"bootloader" blob** — the layer-1 root of trust for a Hook's UI. A wallet renders the on-chain stub, which fetches + **hash-verifies** a larger stage-2 app before running it (secure-boot, rooted in the ledger). The node only **stores + serves** the bytes; the verify/render/sandbox runtime is wallet-side.

Modeled on the **Remarks** amendment.

### What's here (v0)
- `featurePWABoot` (DefaultNo)
- `sfBootBlob` (VL, code 33)
- `ttBOOT_SET` (105) / `SetBoot` transactor
- `sfBootBlob` as an optional field on `AccountRoot`
- `SetBoot`: amendment-gated; **set / replace / delete** the sender's own boot blob; fee = base + 1 drop/byte; 4 KB cap. Per-account, owner-key-mutable.

### Tests
`ripple.app.SetBoot` — 3 cases / 53 checks, green: amendment-disabled → `temDISABLED`; set/replace/delete round-trip on AccountRoot; empty + oversize → `temMALFORMED`, at-cap ok. (`SetHookTSH_test` gets a `testSetBootTSH` stub; full TSH coverage is a follow-up.)

### Open for discussion (#759)
- per-account (here) vs per-hook keying; field-on-AccountRoot (here) vs a dedicated `ltBOOT` object
- immutability (`tfBootImmutable`) — fixed-pin vs owner-key rotation
- size cap + reserve economics (4 KB placeholder)
- the wallet **resolve → verify → render → capability-sandbox** standard
- an `xahc-prover` proof of the bootloader's verify core (the "loads-only-verified" invariant) (scoped: proves the gate's accept logic — `accept ⟹ candidate hash == pin`. The on-chain SetBoot stores the blob verbatim and verifies nothing; the wallet's hashing + stage-2 sandbox stay trusted. Not an end-to-end "whole app proven" claim.)

Marking **draft** — opening early so the ledger shape can be shaped with @RichardAH / the wallet side before it's polished.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RichardAH:
Sorry going a slightly different direction with this and hanging it off AccountSet: https://github.com/Xahau/xahaud/pull/793

PR #766, opened 2026-07-01
fixHookOnV2InstallUpdate Amendment

tequdev:
### Type of Change

### API Impact


src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -1627,10 +1653,21 @@ SetHook::setHook()
...
+                    if ((!newHookOnOutgoing && newHookOnIncoming) ||
+                        (newHookOnOutgoing && !newHookOnIncoming))
+                        return tecINTERNAL;  // LCOV_EXCL_LINE
+                }
+
                 // set the hookon field if it differs from definition
                 if (newHookOn)
                 {
-                    if (*defHookOn == *newHookOn)
+                    if (defHookOn.has_value() && *defHookOn == *newHookOn)
```
shortthefomo: Behaviour change to hook updates applies even on networks where the new rule has not been switched on
The check for whether a stored hook's trigger setting matches its template is changed unconditionally (defHookOn.has_value() at SetHook.cpp:1670) instead of only when the new rule is active, so two servers running different code can disagree on the result of the same update request. Impact: On a network where the new rule is not yet switched on, servers running this build can compute a different hook record than older servers for the same transaction, risking disagreement between servers.

src/test/app/SetHook_test.cpp
```diff
@@ -3005,33 +3256,42 @@ class SetHook0_test : public beast::unit_test::suite
     testInferHookSetOperation()
     {
         testcase("Test operation inference");
+        using namespace jtx;
+
+        Env env{*this};
+
+        SetHookCtx shCtx{
```
shortthefomo, line 3266: Test dereferences a possibly-null transaction pointer when building SetHookCtx
SetHookCtx holds STTx const& tx, and the test initialises it with *env.tx(). Env::tx() (Env.cpp:468-472) returns current()->txRead(txid_).first, which is null when no transaction has been submitted through this Env — which is the case here (a fresh Env is constructed immediately before, and again inside the loop at SetHook_test.cpp:3327-3332). Dereferencing a null shared_ptr to bind the reference is undefined behaviour; it happens to be harmless today only because inferOperation never touches ctx.tx (the HS_ACC() logging macro that reads ctx.tx is only evaluated in trace-enabled log paths). Consider submitting a dummy transaction first, or restructuring so a real STTx is available.

src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -363,6 +423,13 @@ SetHook::validateHookSetEntry(SetHookCtx& ctx, STObject const& hookSetObj)
             // hookon may be present if the user so chooses
             // flags may be present if the user so chooses
 
+            if (ctx.rules.enabled(fixHookOnV2InstallUpdate) &&
```
shortthefomo, line 426: New rule also forbids equal incoming/outgoing masks on install and update
validateHookOn (extracted from the create path) rejects a hook whose sfHookOnIncoming equals sfHookOnOutgoing. Now that it is also invoked for hsoINSTALL and hsoUPDATE under fixHookOnV2InstallUpdate, install/update transactions that set both direction masks to the same value — previously accepted — will return temMALFORMED. The same call also newly rejects install/update transactions that carry direction masks while featureHookOnV2 is off (previously silently accepted and, for install, silently partially applied). Both are plausible intended tightenings, but they are behaviour changes beyond "HookOn vs direction-mask conflicts" and are worth confirming against the amendment's documented scope.

src/xrpld/app/tx/detail/SetHook.cpp
tequdev: fixed

src/test/app/SetHook_test.cpp
tequdev: InferOperation doesn't actually use the transaction data (ctx.tx), but fixed it just in case.

src/xrpld/app/tx/detail/SetHook.cpp
tequdev: This PR is indeed intended to tighten those validation checks.

src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -216,7 +216,12 @@ SetHook::inferOperation(STObject const& hookSetObj)
     bool hasHash = hookSetObj.isFieldPresent(sfHookHash);
     bool hasCode = hookSetObj.isFieldPresent(sfCreateCode);
 
-    if (hasHash && hasCode)  // Both HookHash and CreateCode: invalid
+    bool invalidHookOn = ctx.rules.enabled(fixHookOnV2InstallUpdate) &&
+        hookSetObj.isFieldPresent(sfHookOnOutgoing) ^
+            hookSetObj.isFieldPresent(sfHookOnIncoming);
```
RichardAH, line 221: Mixed bitwise and logical boolean operators are a bit on the nose, wouldn't this work just as well:
`ctx.rules.enabled(fixHookOnV2InstallUpdate) &&
    (hookSetObj.isFieldPresent(sfHookOnOutgoing) != hookSetObj.isFieldPresent(sfHookOnIncoming))`

tequdev: yeah, it will work, I'll fix it

src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -456,56 +530,8 @@ SetHook::validateHookSetEntry(SetHookCtx& ctx, STObject const& hookSetObj)
             }
 
             // validate sfHookOn
-            if (!hookSetObj.isFieldPresent(sfHookOn))
-            {
-                if (!ctx.rules.enabled(featureHookOnV2))
```
RichardAH, line 461: should we just retire the never-enabled featureHookOnV2 and change this amendment to featureHookOnV3?

src/xrpld/app/tx/detail/SetHook.cpp
tequdev: https://github.com/Xahau/xahaud/pull/766/commits/012398bfdcde975d738de90f8966ad1f7c72d0c8

src/xrpld/app/tx/detail/SetHook.cpp
tequdev: Using this fix plus V2 might be fine, but even if that's not the case, V2_1 might still be better.

tequdev: When replacing V2 with V3 (or V2_1), 

it is unlikely to happen, but when a new release comes out and over 80% of the old version reaches consensus on (old) V2, validators on the new version might not be able to reach consensus.

RichardAH: just replace it with V2_1. xahaud validators and stock nodes update pretty quickly and regularly. change v2 to supported::no. No one is going to vote it in as it is so it'll be fine. Better than having a vote-in sequence.

PR #772, opened 2026-08-01
Refactor new account seqno

tequdev:

When creating accounts in GenesisLedger, we previously set the account Sequence to 0. However, since this conflicts with the PseudoAccount requirements, we are changing it to 1. There will be no impact on networks that are already running, and for future networks, there won't be any impact unless you create accounts via GenesisLedger. This specifically addresses an issue that comes up during unittests.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/xrpld/app/tx/detail/Change.cpp
```diff
@@ -554,8 +554,7 @@ Change::activateXahauGenesis()
         {
             sle = std::make_shared<SLE>(kl);
             sle->setAccountID(sfAccount, accid);
-            std::uint32_t const seqno{
-                sb.info().parentCloseTime.time_since_epoch().count()};
+            std::uint32_t const seqno = newAccountSeqNo(sb);
```
shortthefomo, line 557: Accounts funded when a new network is first set up get a different starting counter than before
The starting counter for accounts created during the initial network setup is now computed by a shared helper (newAccountSeqNo(sb) at Change.cpp:557) that only uses the ledger's parent close time when the network feature is already switched on, which is not the case while the setup step itself is running, so those accounts get a different starting counter than the old code produced. Impact: Newly launched networks (and standalone/test networks) will assign different account numbering at bootstrap than before, silently changing ledger contents at network genesis.

src/xrpld/ledger/View.h
```diff
@@ -1266,6 +1266,24 @@ deleteAMMTrustLine(
     std::optional<AccountID> const& ammAccountID,
     beast::Journal j);
 
+[[nodiscard]] inline std::uint32_t
```
shortthefomo, line 1269: Starting counter for brand-new accounts changes without an amendment guard
The starting counter given to newly created accounts is now silently bumped from zero to one when the ledger's parent close time is zero (newAccountSeqNo at View.h:1269-1285) with no amendment gating the change, so nodes running different versions of the software could build different ledgers. Impact: A transaction-processing behavior change ships unguarded, which risks divergence between old and new nodes on any network where the condition can occur.

src/test/app/AMMExtended_test.cpp
```diff
@@ -150,12 +150,15 @@ struct AMMExtended_test : public jtx::AMMTest
 
             env.fund(XRP(20'000), alice, bob, carol, gw1, gw2);
             env.fund(XRP(20'000), dan);
+            env.close();
```
shortthefomo, line 153: Extra env.close() calls added to tests may mask ordering-sensitivity rather than fix it
env.close() calls were inserted after funding/trust/payment batches in AMMExtended_test.cpp:153-161 and URIToken_test.cpp:992. These make the tests pass again after the sequence-number shift, but they also change what is being exercised (transactions now land in separate ledgers). 

Worth confirming that the original single-ledger flow is still covered elsewhere, since the closes were added purely to stabilize expectations.

src/xrpld/app/tx/detail/Change.cpp
tequdev: This process is not executed on currently running networks, so there is no impact.

Also, I had added a comment regarding this inside the newAccountSeqNo() function.

src/xrpld/ledger/View.h
tequdev: Since time=0 is a value that can only occur in the genesis ledger, and it is impossible for an active network to reach the genesis ledger again, an Amendment Guard is not necessary.

src/test/app/AMMExtended_test.cpp
tequdev: It is not intended to run fund() and the subsequent transactions in the same ledger.

Inserting close() here should be fine.

PR #776, opened 2026-08-04
fixHookNameValidation Amendment

tequdev:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/xrpld/app/tx/detail/Transactor.cpp
+++ b/src/xrpld/app/tx/detail/Transactor.cpp
@@ -155,7 +155,11 @@ preflight1(PreflightContext const& ctx)
             !ctx.rules.enabled(featureNamedHooks))
             return temMALFORMED;
 
-        if (!SetHook::validateHookName(ctx.tx.getFieldVL(sfHookName), ctx.j))
+        auto const& name = ctx.tx.getFieldVL(sfHookName);
+
+        if (name.size() == 0 && ctx.rules.enabled(fixHookNameValidation))
+            return temMALFORMED;
+        if (!SetHook::validateHookName(name, ctx.j))
             return temMALFORMED;
     }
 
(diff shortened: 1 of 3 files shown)


src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -155,7 +155,11 @@ preflight1(PreflightContext const& ctx)
             !ctx.rules.enabled(featureNamedHooks))
             return temMALFORMED;
 
-        if (!SetHook::validateHookName(ctx.tx.getFieldVL(sfHookName), ctx.j))
+        auto const& name = ctx.tx.getFieldVL(sfHookName);
```
shortthefomo, line 158: Empty HookName is still accepted inside SetHook hook objects
The new amendment-gated rejection of a zero-length sfHookName is applied only to the transaction-level field in preflight1. SetHook::validateHookName (SetHook.cpp:625-638) explicitly permits name.size() == 0, and the SetHook hook-object validation path (SetHook.cpp:517-522) is unchanged, so a hook can still be installed with an empty sfHookName field. Such a hook is then treated as unnamed by the execution/fee paths (Transactor.cpp:291-293 and :1362-1364 only set requiredHookName when size > 0), i.e. an empty name field is stored but semantically ignored. If the intent of the amendment is that an empty name is malformed, consider whether the same rule should apply to the SetHook side for consistency.

tequdev: HookName in SetHook represents the deletion of a field.

You cannot set a field with size==0.

```cpp
                    if (newHookName->size() == 0)
                    {
                        newHook.makeFieldAbsent(sfHookName);
                    }
                    else
                    {
                        newHook.setFieldVL(sfHookName, *newHookName);
                    }
```

PR #777, opened 2026-08-05
fix(manifest): bound untrusted cache without stalling gossip

sublimator:
## Summary

- Caps evictable manifest retention at 1,000 identities. Once full, eviction permits allow a burst of 10 replacements and refill roughly once per second; eviction prefers identities without current validations, then falls back to uniform random selection.
- Adds an optional `candidates` tier to signed publisher lists for zero-weight validator monitoring. Candidate membership never contributes listing or trust. Each generation and the deterministically selected current candidate set are capped at 1,000 identities.
- Protects local, tier-1, and selected tier-2 identities from eviction while keeping `ManifestCache` as the sole sequence, revocation, and signing-key-to-master authority. Provenance controls retention and relay independently of consensus eligibility.
- Relays accepted manifests for protected identities and accepted updates for retained evictable identities without using the shared `HashRouter`. First-seen valid evictable identities are retained up to the cap but do not relay.
- Bounds inbound manifests by encoded size, wire and uncompressed frame size, a 1,000-entry frame cap, and a 200-entry processing cap for unprotected manifests.
- Sends protected manifests to new peers in entry- and byte-bounded frames, followed by at most 200 evictable manifests.

## Lifecycle

- Removing an identity from every protected source makes its cached manifest evictable; it is retained if capacity permits and otherwise discarded.
- Embedded manifests in current, available publisher lists establish tier-1/tier-2 baselines; bare entries establish protection only. Newer gossip can advance retained manifests while running.
- Persisted validator manifests load provisionally protected, then startup reconciliation reclassifies DB-only rows as evictable. Shutdown save filtering is unchanged: terminal revocations are saved, while non-revocations are saved only for listed validators.

### DIFF
--- a/src/test/consensus/Validations_test.cpp
+++ b/src/test/consensus/Validations_test.cpp
@@ -544,6 +544,49 @@ class Validations_test : public beast::unit_test::suite
     }
 
     void
+    testGetCurrentPublicKeys()
+    {
+        using namespace std::chrono_literals;
+        testcase("Current public keys");
+
+        LedgerHistoryHelper h;
+        Ledger ledgerA = h["a"];
+        Ledger ledgerAC = h["ac"];
+
+        TestHarness harness(h.oracle);
+        Node a = harness.makeNode(), b = harness.makeNode();
+        b.untrust();
+
+        for (auto const& node : {a, b})
+            BEAST_EXPECT(
+                ValStatus::current == harness.add(node.validate(ledgerA)));
+
+        {
+            hash_set<PeerID> const expectedKeys = {a.nodeID(), b.nodeID()};
+            BEAST_EXPECT(harness.vals().getCurrentNodeIDs() == expectedKeys);
+        }
+
+        harness.clock().advance(3s);
+
+        // Change keys and issue partials
+        a.advanceKey();
+        b.advanceKey();
+
+        for (auto const& node : {a, b})
+            BEAST_EXPECT(
+                ValStatus::current == harness.add(node.partial(ledgerAC)));
+
+        {
+            hash_set<PeerID> const expectedKeys = {a.nodeID(), b.nodeID()};
+            BEAST_EXPECT(harness.vals().getCurrentNodeIDs() == expectedKeys);
+        }
+
+        // Pass enough time for them to go stale
+        harness.clock().advance(harness.parms().validationCURRENT_LOCAL);
+        BEAST_EXPECT(harness.vals().getCurrentNodeIDs().empty());
+    }
+
+    void
     testTrustedByLedgerFunctions()
     {
         // Test the Validations functions that calculate a value by ledger ID
@@ -1089,6 +1132,7 @@ class Validations_test : public beast::unit_test::suite
         testOnStale();
         testGetNodesAfter();
         testCurrentTrusted();
+        testGetCurrentPublicKeys();
         testTrustedByLedgerFunctions();
         testExpire();
         testFlush();
--- a/src/test/consensus/Validations_test.cpp
+++ b/src/test/consensus/Validations_test.cpp
@@ -544,58 +544,6 @@ class Validations_test : public beast::unit_test::suite
     }
 
     void
-    testGetCurrentPublicKeys()
-    {
-        using namespace std::chrono_literals;
-        testcase("Current public keys");
-
-        LedgerHistoryHelper h;
-        Ledger ledgerA = h["a"];
-        Ledger ledgerAC = h["ac"];
-
-        TestHarness harness(h.oracle);
-        Node a = harness.makeNode(), b = harness.makeNode();
-        b.untrust();
-
-        for (auto const& node : {a, b})
-            BEAST_EXPECT(
-                ValStatus::current == harness.add(node.validate(ledgerA)));
-
-        {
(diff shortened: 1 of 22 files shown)


src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -1955,20 +2244,21 @@ ValidatorList::updateTrusted(
...
-                // get processed in. The expiration check below will do the rest
-                // of the work
-                if (current.validUntil <= closeTime)
-                    current.list.clear();
 
-                updatePublisherList(pubKey, current, oldList, lock);
+                bool const expired = current.validUntil <= closeTime;
+                collection.status = expired ? PublisherStatus::expired
+                                            : PublisherStatus::available;
+                auto const currentMasters = expired ? std::vector<PublicKey>{}
```
shortthefomo: Validators can be silently dropped from the trusted set after a publisher list arrives already out of date
When a queued validator list becomes current after its own expiry date, the validators it names are kept in the stored list but are never counted as listed (updatePublisherList receives an empty master set at ValidatorList.cpp:2256-2259), so a later refresh from the same publisher subtracts counts that were never added and can remove validators that other sources still vouch for. Impact: Validators may lose their trusted status even while they are still published by other sources, shrinking the trusted set and disturbing quorum; the node also stops flagging that it has no usable validator list in this case.

src/xrpld/overlay/detail/OverlayImpl.cpp
```diff
@@ -679,9 +754,22 @@ OverlayImpl::onManifests(
...
+        JLOG(journal.warn())
+            << "Manifests: message had " << total
+            << " entries; processed all trusted plus the first "
+            << kMaxManifestsPerMessage << " untrusted";
+    }
+
     if (!relay.list().empty())
-        for_each([m2 = std::make_shared<Message>(relay, protocol::mtMANIFESTS)](
-                     std::shared_ptr<PeerImp>&& p) { p->send(m2); });
+        for_each([m2 = std::make_shared<Message>(relay, protocol::mtMANIFESTS),
```
shortthefomo, line 770: Outgoing relay bundle is entry-bounded but not byte-bounded
relay accumulates every accepted manifest from one incoming message: up to kMaxManifestsPerMessage (200) untrusted rotations plus an unbounded number of relay-eligible (listed/candidate) entries. The snapshot path in OverlayImpl::getManifestsMessages chunks by both entry count and maximumManifestsMessageSize, but this relay path does neither. 200 max-size entries (~364 bytes each with protobuf framing) come to ~72.8 KB, just under the 73.2 KB receive limit, so a single additional relay-eligible entry pushes the bundle over the limit and every patched peer will silently drop the whole relay. Applying the same chunking helper here would make the send and receive limits consistent.

src/test/consensus/Validations_test.cpp
```diff
@@ -543,49 +543,6 @@ class Validations_test : public beast::unit_test::suite
         BEAST_EXPECT(harness.vals().currentTrusted().empty());
     }
 
-    void
```
shortthefomo, line 546: Unrelated deletion of a consensus validations test
testGetCurrentPublicKeys is removed from Validations_test.cpp along with its call in run(), but nothing in this PR touches Validations or getCurrentNodeIDs (still exercised only indirectly at Validations_test.cpp:1046). This looks like an accidental deletion that reduces coverage of key rotation for current validations; consider restoring it.

src/xrpld/app/misc/detail/ValidatorList.cpp
sublimator: Thanks @shortthefomo

src/xrpld/overlay/detail/OverlayImpl.cpp
sublimator: Thanks @shortthefomo

src/test/consensus/Validations_test.cpp
sublimator: Thanks @shortthefomo

src/xrpld/app/misc/detail/ValidatorList.cpp
sublimator: Fixed in 8bd8d9a12. An already-expired pending list now clears its never-counted parsed membership before following the ordinary available-to-expired path. That prevents a later refresh from decrementing another publisher's contribution and preserves the existing UNL-blocked signal. Added a cross-publisher regression.

src/xrpld/overlay/detail/OverlayImpl.cpp
sublimator: I traced this through the receive path. onManifests is only invoked for a frame whose wire and uncompressed payload sizes have already passed maximumManifestsMessageSize in ProtocolMessage; the live relay is an exact-byte subset of those same entries and adds nothing. It therefore cannot exceed the admitted source frame. Snapshot construction is different because it aggregates cache entries, so its byte/entry chunking remains necessary. No code change here.

src/test/consensus/Validations_test.cpp
sublimator: Restored in 8bd8d9a12. The production API removal was intentional, but removing this existing getCurrentNodeIDs rotation/staleness coverage was not.

PR #778, opened 2026-08-07
fixHookAPISType Amendment

tequdev:

  - Adds the `fixHookAPISType` amendment.
  - Prevents the Hook API PathSet parser from reading past the serialized input boundary

### Context of Change

The original requested fix is the PathSet boundary check in the first commit. The follow-up commits extend the same amendment to related serialized-object handling:

  - Reject truncated fixed-length, `Issue`, and `XChainBridge` field payloads before they are read.
  - Require `sto_emplace` source objects to be fully valid.
  - Require injected `field_object` data to contain exactly one complete field.
  - Validate complete inputs in `sto_subfield` and `sto_subarray` before returning an early match.
  - Require wrapped arrays passed to `sto_subarray` to end with the array marker (`0xF1`).

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/test/app/HookAPI_test.cpp
+++ b/src/test/app/HookAPI_test.cpp
@@ -3941,310 +3941,113 @@ public:
 
         auto const _source_object =
             *strUnHex("81140000000000000000000000000000000000000000");
+
+        std::vector<std::pair<uint32_t, std::string>> const data_list = {
+            {// UINT16 Version
+             sfVersion.getCode(),
+             "10100001"},
+            {// UINT32 Sequence
+             sfSequence.getCode(),
+             "2400000001"},
+            // UINT64
+            {// Uint64 ExchangeRate
+             sfExchangeRate.getCode(),
+             "360000000000000001"},
+            {// UINT256 EmailHash
+             sfEmailHash.getCode(),
+             "4100000000000000000000000000000000"},
+            {// UINT256 ObjectID
+             sfObjectID.getCode(),
+             "5E000000000000000000000000000000000000000000000000000000000000000"
+             "0"},
+            {// Amount NativeAmount
+             sfAmount.getCode(),
+             "61999999999999999999999999999999999999999999999999999999999999999"
+             "999999999999999999999999999999999"},
+            {// Amount IOUAmount
+             sfAmount.getCode(),
+             "614999999999999999"},
+            {// Object AmountEntry
+             sfAmountEntry.getCode(),
+             "E05B614000000000000064E1"},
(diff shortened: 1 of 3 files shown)


src/xrpld/app/hook/detail/HookAPI.cpp
```diff
@@ -3066,6 +3096,9 @@ HookAPI::get_stobject_length(
                 if (flag & 0x20)  // issuer
                     length += 20;
 
+                if (rules.enabled(fixHookAPISType) && upto + length > end)
```
sublimator, line 3099: `>=` ?

shortthefomo: The new boundary guard for path data allows reading exactly at the end of the buffer (upto + length > end) before the very next line reads that position, so a truncated path still causes a read of one byte outside the supplied data. Impact: A hook can hand the node truncated path data and make it read memory it does not own, which can crash the node under hardened/instrumented builds.

src/test/app/HookAPI_test.cpp
```diff
@@ -3941,310 +3941,119 @@ class HookAPI_test : public beast::unit_test::suite
...
-            source_object.insert(
-                source_object.end(), field_object.begin(), field_object.end());
-            BEAST_EXPECT(result.value() == source_object);
-
-            auto const erase_result = api.sto_emplace(
-                result.value(), std::nullopt, sfTickSize.getCode());
-            BEAST_EXPECT(erase_result.has_value());
-            BEAST_EXPECT(erase_result.value().size() == _source_object.size());
-            BEAST_EXPECT(erase_result.value() == _source_object);
+            returnError = !result1.has_value() && !result2.has_value() &&
```
shortthefomo, line 4022: Negative-case loop only asserts the last data entry
Inside the loop over data_list, returnError is assigned (=) rather than accumulated on each iteration, so after the loop only the results for the final entry (sfBaseAsset) are reflected. The subsequent BEAST_EXPECT(returnError) / BEAST_EXPECT(!returnError) therefore silently ignores all earlier types (Version, Sequence, Amount, PathSet, XChainBridge, ...), which are precisely the interesting cases for this amendment. Accumulating with &= per iteration (or asserting inside the loop with BEAST_EXPECTS(..., data.second)) would give real coverage.
Also note that in the non-amendment branch the test asserts !returnError, which is satisfied as soon as one of the three malformed variants errors out — it does not verify the pre-amendment behavior meaningfully.

PR #779, opened 2026-08-10
Onchain manifests

RichardAH:
# OnChainManifests
 
Adds manifests to ledger storage as a third source of truth, in addition to the VL and the gossip protocol. A node derives the current signing key for any master key from ledger state alone, without having been online when the rotation was gossiped. Gossip is capped and prioritised to important manifests only.
 
| | |
|---|---|
| **Amendment** | `OnChainManifests` — `Supported::yes`, `VoteBehavior::DefaultNo` |
| **Transaction** | `SetManifest` / `ttMANIFEST_SET` (91). One field: `sfManifest` |
| **Ledger object** | `Manifest` / `ltMANIFEST` (`0x004D`), RPC name `manifest_entry` |
| **New fields** | `sfManifest` (OBJECT 90), `sfManifestID` (UINT256 91, on `AccountRoot`) |
| **New TERs** | `tefPAST_MANIFEST_SEQ`, `tefREVOKED_MANIFEST` |
| **Keylet** | `keylet::manifest(pk)` — stored twice, under master key and signing key |
| **Fee** | Base + 1 drop per manifest byte, capped at +20%. Paid by the master key's account |
| **Signature** | None on the envelope. The manifest's own signatures are the authority |
| **Gossip** | Pinned set (trusted validators) plus up to 64 more, LRU ranked |
 
Submission is not restricted to the validator. Any node holding a manifest can submit it, and the master key's account pays the fee regardless of who submits.
 
## Getting the manifest in hex
 
From `validator-keys`, if you have the key file. Undocumented in the tool's guide but present in the source and in `--help`:
 
```
$ validator-keys show_manifest hex
Manifest #1 (Hex):
24000000017121ED26F62287F876...
```
 
Otherwise convert the `[validator_token]` from the config. It is base64 JSON with the manifest as a base64 field:
 
```bash
echo "$TOKEN" | tr -d '[:space:]' | base64 -d \
  | sed 's/.*"manifest":"\([^"]*\)".*/\1/' | base64 -d \
  | od -An -tx1 | tr -d ' \n' | tr a-f A-F
```
 
A v0 manifest starts `24` (`sfSequence`); one carrying `sfVersion` starts `1010`. In base64 that is `JAAAAAF...`.
 
## Submitting
 
```json
{"command": "submit", "manifest": "24000000017121ED26F62287F876..."}
```
 
The server builds, prices and submits the envelope. Response is the ordinary `engine_result` / `tx_json`. `manifest` and `tx_blob` are mutually exclusive.
 
| Condition | Response |
|---|---|
| Amendment not enabled | `notEnabled` |
| Both `manifest` and `tx_blob` | `invalidParams` |
| Bad hex, or empty | `invalidParams` |
| Not a manifest, or signatures don't verify | `invalidManifest` |
 
## Reading back
 
The existing `manifest` RPC returns on-ledger manifests, because `ManifestCache` is seeded from the ledger at startup:
 
```json
{"command": "manifest", "public_key": "nHU..."}
```
 
`account_objects` with `"type": "manifest_entry"` lists them on an account. `ledger_entry` has no `manifest` option — use the raw `index`.
 
## Envelope constraints
 
The transaction carries no account signature. `checkValidity()` verifies the manifest's master and ephemeral signatures instead, gated on the amendment. Before activation a `SetManifest` is an unsigned transaction and is rejected as one.
 
The sender is therefore unauthenticated, so every field a relayer could otherwise choose is pinned. `sfSequence` must be 0, `sfSigningPubKey` and `sfTxnSignature` empty, `sfAccountTxnID` / `sfTicketSequence` / `sfSigners` absent, and `sfAccount` must be the master key's account. Anything else is `temMALFORMED`.
 
`sfFee` cannot be bounded in preflight because the base fee is not in scope until preclaim. `checkFee()` caps it at base + 20%. Without a cap, an observer of a not-yet-recorded manifest could wrap it with `sfFee` set to the validator's entire balance. The band leaves headroom for a fee floor that rose after the transaction was built and bounds the loss to 20%.
 
Replay is covered twice. A byte-identical resubmit is `tefALREADY`; a differently priced one fails the strictly-increasing sequence test as `tefPAST_MANIFEST_SEQ`. Both are `tef`, so a replay is never included and never claims a fee. A revoked master key is closed permanently with `tefREVOKED_MANIFEST`. A manifest naming another validator's key as its ephemeral key is `tecDUPLICATE`, the on-chain equivalent of `ManifestCache`'s `badEphemeralKey` check.
 
## Ledger object
 
Stored twice, at `keylet::manifest(masterKey)` and `keylet::manifest(signingKey)`, cross-linked by `sfManifestID` and referenced from the account root. Either key resolves in one hop. A revocation has no signing key and stores one copy.
 
The object mirrors the manifest's `sfSignature` and `sfMasterSignature`, and omits `sfVersion` where the manifest did, so the signed bytes reconstruct exactly. Any node reading it can re-verify it and re-serve it to peers.
 
## Node behaviour
 
- `ManifestCache::applyLedger()` seeds the cache from ledger state at startup, recovering rotations missed while offline.
- Validators auto-publish. Each ledger, `publishNewerManifests()` compares held manifests against the ledger and submits anything newer.
- Manifests bypass the transaction queue and fee escalation. `sfSequence` 0 can never match an account sequence, so queueing would reject rather than hold them.
- The ephemeral key's logical account is a weak TSH — observe, not rollback. Usually no such account exists and nothing fires.
- Hooks get `ttMANIFEST_SET`, `sfManifest` and `sfManifestID` in `tts.h` and `sfcodes.h`.
## Before activation
 
`submit` refuses the `manifest` parameter. A hand-built `SetManifest` fails signature checking like any unsigned transaction. No existing transaction, ledger object or RPC response shape changes.

## Related research branches

- #777 — bounded cache and ingress controls
- #794 — validation-driven manifest propagation
- #795 — combined propagation, durability, and ownership


src/xrpld/app/misc/NetworkOPs.cpp
```diff
@@ -1874,13 +1946,34 @@ NetworkOPsImp::beginConsensus(
...
         app_.validators().setNegativeUNL(prevLedger->negativeUNL());
+    // Pull in any manifests published on-ledger before the trusted set is
+    // recomputed, so a validator that rotated its ephemeral key on-chain is
+    // resolved to the new signing key in this same round. The master keys come
+    // from the published lists, so this needs no bootstrap: only the ephemeral
+    // half of the mapping ever comes from a manifest.
+    if (prevLedger->rules().enabled(featureOnChainManifests))
+    {
+        app_.validatorManifests().applyLedger(
+            *prevLedger, app_.validators().getTrustedMasterKeys());
```
sublimator, line 1957: This only ingests manifests for already-trusted keys. A cache miss for a non-UNL signing key never consults its on-ledger `keylet::manifest`, so self-submission does not make the manifest discoverable. Can the resolver fall back to the validated ledger?

RichardAH: significant oversight indeed, I think the following commit patches

sublimator (approved):
Looks good as the on-ledger manifest foundation. Needs a follow-up before amendment activation for bounded ingress and cache retention.

PR #782, opened 2026-08-12
feat(hooks): add JSHookz QuickJS integration seam

sublimator:
## Summary

Draft integration for [JSHookz](https://github.com/sublimator/jshookz).

JSHookz compiles QuickJS to WebAssembly so JavaScript/TypeScript Hooks can be
deterministically metered. Xahaud then translates the provider's small
primitive import surface to the existing host functions used by C Hooks.

This branch is primarily a proof of the integration shape, not a
production-ready activation.

## What this proves

- Ordinary C Hooks continue to validate and execute through WasmEdge.
- QuickJS executes through Wasmtime, using the same visible host-operation
  bodies in `applyHook.cpp`.
- The neutral catalogue mechanically exposes all 75 existing Hook operations
  without a second implementation of their ledger semantics.
- The current immutable QuickJS v1 policy selects and meters 13 operations;
  that selection is provider policy, not an architectural limit.
- Sealed provider identity, bytecode ABI, runtime profile, admission,
  execution, state commit/rollback, callbacks, emissions, memory growth,
  host-work charging, concurrency, and exact fuel receipts are exercised end
  to end.
- The provider snapshot and embedded Env fixtures can be refreshed
  mechanically with `bin/update-jshookz-snapshot`.

The intended target is Wasmtime because its AOT and high-performance execution
are useful for the relatively large compiled QuickJS provider. The required
Xahaud changes are smaller than they first appear: a generic Wasmtime callback,
a mechanically generated descriptor catalogue, provider/profile plumbing, and
the existing shared host bodies.

## Outstanding

- [ ] Define and review production activation and retained provider
      registration.
- [ ] Finalize fees, metering economics, and the on-ledger meaning of QuickJS
      execution cost.
- [ ] Retain genuinely versioned host implementations before activation so
      historical profiles replay immutably.
- [ ] Benchmark interpreter/AOT startup, steady-state execution, memory,
      provider size, and representative C-vs-JS Hooks.
- [ ] Continue the public TypeScript API and audited-port ratchet in JSHookz.
- [ ] Decide how provider artifacts are packaged and supplied by release
      builders.
- [ ] Complete production deployment/operator documentation.

## Verification

The branch includes the authoritative provider-backed Xahau Env workflow plus
focused C-Hook/host-policy/memory tests. JSHookz's own build verifies the sealed
provider, declarations, profiles, codec fixtures, compiler/package flow, and
host harness.


src/xrpld/app/hook/detail/WasmEdgeHookEngine.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+        JLOG(journal.trace())
+            << "HookInfo[" << hookContext.result.account << "-"
+            << hookContext.result.otxnAccount << "]: creating wasm instance";
+        WasmEdge_LogOff();
+
+        WasmEdgeVM engine;
+        if (!engine.sane())
+        {
+            WasmEdge_ModuleInstanceDelete(imports);
```
shortthefomo: Hook execution engine frees its shared function table before shutting down the virtual machine that is still using it
The block of host functions handed to the WebAssembly engine is destroyed (WasmEdge_ModuleInstanceDelete(imports) at WasmEdgeHookEngine.cpp:246) before the engine that was given it is shut down, so every Hook run can touch memory that has already been released.
Impact: Nodes can crash or corrupt memory while running any Hook, on every transaction that triggers one.

src/xrpld/app/tx/detail/SetHook.cpp
```diff
@@ -58,6 +62,64 @@ namespace ripple {
...
+            return declaredApiVersion == 0
+                ? ArtifactInstallability::allowed
+                : ArtifactInstallability::apiMismatch;
+
+        case hook::artifact::Kind::quickJSBytecode:
+            if (declaredApiVersion != 1)
+                return ArtifactInstallability::apiMismatch;
+            if (!rules.enabled(featureJSHooks))
+                return ArtifactInstallability::amendmentDisabled;
+#ifdef ENABLE_TESTS
```
shortthefomo: Whether a JavaScript Hook can be installed depends on whether the binary was compiled with unit tests enabled
Acceptance of a QuickJS Hook deployment is decided by a compile-time test switch (#ifdef ENABLE_TESTS at SetHook.cpp:95-98), so two servers built from the same source can disagree about whether the very same transaction is valid.
Impact: If the JavaScript Hooks feature is ever switched on, test-enabled and release servers would accept different transactions and the network could split.

src/xrpld/app/hook/detail/WasmEdgeHookEngine.cpp
sublimator: Thanks !

src/xrpld/app/tx/detail/SetHook.cpp
sublimator: Thanks!

PR #799, opened 2026-09-07
fix(manifest): bound ingress and restrict gossip to local lists

sublimator:


src/xrpld/app/misc/detail/Manifest.cpp
```diff
@@ -404,6 +404,31 @@ ManifestCache::touch(PublicKey const& masterKey) const
...
+            oldest = used;
+        }
+    }
+    if (victim == evictable_.end())
+        return;
+    auto const row = map_.find(*victim);
+    if (row->second.signingKey)
+        signingToMasterKeys_.erase(*row->second.signingKey);
+    lastUsed_.erase(*victim);
+    map_.erase(row);
```
sublimator: This question is superseded by the pending update: gossip is limited to locally listed/configured masters, and new unlisted identities are discovered through the ledger. No general eviction in this PR; a small ledger-only LRU remains a separate follow-up.

sublimator:
~~TODO: we need to account for ricochet, currently the manifests are relayed upon entry into a cache, which is now evictable~~

src/xrpld/overlay/detail/PeerImp.cpp
```diff
@@ -1057,12 +1057,49 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMManifests> const& m)
...
 
+    std::vector<Manifest> candidates;
+    int cost = 0;
+    for (auto const& item : m->list())
+    {
+        auto manifest = deserializeManifest(item.stobject());
+        if (!manifest)
+            cost += Resource::feeInvalidData.cost();
+        else if (!app_.validatorManifests().isGossipCandidate(*manifest))
+            cost += Resource::feeUselessData.cost();
```
RichardAH, line 1077: Won't this cause every new build to charge every peer on an old build partitioning the network? We should only introduce something like this later, after on-chain manifests goes live?

PR #806, opened 2026-09-10
feat: add NoRecipientLimit amendment

sublimator:
Today a Payment can fail because the recipient has reached its trust-line limit, even though Remit already lets that limit be exceeded.

This change lets an account receive more of the named issuer's token on an existing trust line, whoever pays: the exempt step is the issuer's credit into the destination, and a holder's payment that ripples through the issuer ends with that same step. Limits still apply to accounts used as intermediaries along the payment path. Funds, authorization and freeze checks are unchanged.

A companion change keeps Remit-created trust lines when their balance returns to zero, retaining their owner reserve.

## Scope

Against `c208a40ce`: 125 source lines added, 13 removed, across 11 files; 499 test lines added, 53 removed, across 9 files.

Of the 125 added source lines, **77 are code**, 41 are comments and 7 are blank.

| Area | Files | Lines |
|---|---|---|
| Engine exemption | `DirectStep.cpp`, `Credit.cpp`, `View.h` | +49 / -3 |
| Persist flag | `SetTrust.cpp`, `View.cpp`, `Remit.cpp` | +63 / -7 |
| Protocol: flags, amendment, hook headers | `LedgerFormats.h`, `TxFlags.h`, `features.macro`, `hook/*.h` | +13 / -3 |
| New tests | `Flow_test`, `SetTrust_test`, `Remit_test` | +374 / -11 |
| Gates on existing tests | `Check`, `PayChan`, `Path`, `TrustAndBalance`, `URIToken`, `DeliveredAmount` | +125 / -42 |

## The invariant

A trust-line limit governs an account being used as an intermediary, never an account receiving its issuer's token. The issuer's credit into a transaction's destination ignores the recipient's limit. A hop through a third account inside a payment path obeys that account's limit, because that account is lending its balance sheet to someone else's payment. A non-issuer crediting the destination directly still obeys the destination's limit on that account: the limit is the destination's acceptance of that account's IOU, not a cap on the token it asked for. The same rule keeps a sender from over-redeeming to the issuer by issuing its own IOU on the way out.

Several transfer paths already allow receipt above a trust-line limit: Remit, offer crossing, NFT and URIToken sales, AMM operations, and checks under `CheckCashMakesTrustLine`. This amendment changes the two places still enforcing it: the payment engine's issuer-to-recipient step, and token escrow and payment-channel settlement when someone other than the destination finishes the transfer.

## The one exception

After the amendment, a limit is enforced on a line whose owner is the *recipient* in exactly one case:

| Exception | Example | Why |
|---|---|---|
| The credit is another account's IOU, not the issuer named in the payment | Payment names carol's USD, path ends bob → carol on carol's line with bob: carol's limit on bob still caps it. Same when alice pays gw more USD than she holds: the shortfall would be alice's own IOU onto gw, and gw's limit (0) refuses it. | Your limit on a peer is your acceptance of *that peer's* IOU. You asked for the issuer's token, not theirs. |

Everything else where you are the recipient ignores the limit. Advisory only, not enforcement: `path_find` and `account_currencies` still treat a full line as uncreditable.

## More to think about

- **Clearing the persist bit.** Xaman's remove-token flow sends a TrustSet with limit 0 and the freeze/NoRipple flags ([TokenSettingsOverlay.tsx](https://github.com/XRPL-Labs/Xaman-App/blob/01e2538b08ef/src/screens/Overlay/TokenSettings/TokenSettingsOverlay.tsx#L408-L431)). On a Remit-created line that TrustSet no longer deletes the line, and a persisting line also blocks AccountDelete. One way out: any TrustSet from a side clears that side's persist bit unless it sets `tfSetPersist`, and `tfClearPersist` goes away.
- **Who can reach the exempt step.** Any payer whose strand ends with the issuer's step, which includes a holder rippling through a DefaultRipple gateway. Remit already allows the same. The narrower option is issuer-signed payments only, a one-line predicate change.
- **The name.** `NoRecipientLimit` is a placeholder.
- **Not in this PR.** The pathfinder and `account_currencies` still treat a full destination line as uncreditable, `account_lines` does not report persist, and persist does not veto AMM trust-line teardown.
- **You can't stop your issuer paying you, except by removing the line.** Payment has no "disallow incoming" flag like Remit does. DepositAuth works, but it blocks everything. And if you remove the line, a Remit can put it right back unless you set `DisallowIncomingRemit`.
- **Remit can give you a line; persist makes it stay.** That was already true for the line and its reserve. Now it stays after you spend to zero, until you clear it. The clearing question above decides how easy that is.
- **Payment still doesn't create lines.** Remit already does that job. Payment could too, but it means picking a reserve rule (Remit charges the sender; CashCheck uses the casher's own) and a fair bit more code, so it stays out of this change.
- **One amendment or two?** The engine change and the persist flag could be voted on separately.
- **Wallets won't notice yet.** Xaman checks the recipient's limit itself before sending and pushes you to Remit. Nothing changes for users until the app skips that check for issuer payments.
- **One bit of dead weight later.** `CheckCashMakesTrustLine` raises the casher's limit around the engine. Once this is on, it doesn't need to.

## How Xaman handles this today

When you send a token, Xaman checks the recipient's trust line before it builds anything ([RecipientStep.tsx#L748-L865](https://github.com/XRPL-Labs/Xaman-App/blob/01e2538b08ef/src/screens/Send/Steps/Recipient/RecipientStep.tsx#L748-L865)):

- Skipped when sending a token back to its issuer, e.g. depositing to Bitstamp.
- Otherwise it fetches the recipient's line with the issuer ([LedgerService.ts#L593](https://github.com/XRPL-Labs/Xaman-App/blob/01e2538b08ef/src/services/LedgerService.ts#L593)) and refuses if:
  - there is **no line** (missing, or default 0 limit/balance), or
  - the amount plus the current balance would go over the **limit** ([L775-L778](https://github.com/XRPL-Labs/Xaman-App/blob/01e2538b08ef/src/screens/Send/Steps/Recipient/RecipientStep.tsx#L775-L778)).

On Xahau, the Payment refusal due to a missing or non-supporting line comes with an offer: "Do you want to send your payment as a 'Remit' transaction instead? This will automatically create the Trustline in the recipient's account, but you will pay an extra 0.2 XAH to satisfy their TrustLine reserve" ([en.json#L1219](https://github.com/XRPL-Labs/Xaman-App/blob/01e2538b08ef/src/locale/en.json#L1219)).

Xaman would simply need to update to remove the limit check, as the limit is only applied to intermediaries, not an end recipient. Remit retains its job as the way to pay someone who has no line yet.

## Recipient limit by transaction, before and after

IOU trust-line credits only. "Enforced" means the recipient's limit capped or refused the credit.

| Transaction delivering a token | Before | After |
|---|---|---|
| Payment, issuer's hop into the recipient (direct, holder to holder via the issuer, DEX book then the implied issuer hop) | enforced | ignored |
| Payment, last hop on another account's line (path names the recipient's own currency) | enforced | enforced |
| Payment, recipient used as an intermediary earlier in the same strand | enforced | enforced |
| EscrowFinish / PayChanClaim (IOU), finished by the recipient | ignored | ignored |
| EscrowFinish / PayChanClaim (IOU), finished by anyone else | enforced | ignored |
| CheckCash, `CheckCashMakesTrustLine` on (Xahau today) | ignored | ignored |
| CheckCash, `CheckCashMakesTrustLine` off | enforced | ignored |
| Remit | ignored | ignored |
| NFTokenAcceptOffer (sale, broker fee, issuer transfer fee) | ignored | ignored |
| URIToken buy (sale proceeds to the seller) | ignored | ignored |
| OfferCreate (taker's receipt) | ignored | ignored |
| AMM create / deposit / withdraw / bid, and AMMClawback's withdrawal credit (`AMM` is `Supported::no`) | ignored | ignored |
| XChainBridge IOU claims and commits into a locking door (`Supported::no`) | as Payment | as Payment |
| Clawback / AMMClawback debit; Import, GenesisMint, ClaimReward, MPT | n/a | n/a |

### DIFF
--- a/src/xrpld/app/paths/detail/DirectStep.cpp
+++ b/src/xrpld/app/paths/detail/DirectStep.cpp
@@ -44,7 +44,6 @@ protected:
     // Charge transfer fees when the prev step redeems
     Step const* const prevStep_ = nullptr;
     bool const isLast_;
-    //@@start issues-to-dst-member
     // This step is the delivered asset's issuer crediting the strand's
     // destination. Not the same as isLast_: the implied issuer-to-destination
     // step after a book or AMM is built without isLast, and isLast_ also
@@ -56,7 +55,6 @@ protected:
     // (its own USD line feeding a USD/EUR book that delivers EUR), and that
     // hop is capped like any intermediary hop.
     bool const issuesToDst_;
-    //@@end issues-to-dst-member
     beast::Journal const j_;
 
     struct Cache
@@ -112,11 +110,9 @@ public:
         , currency_(c)
         , prevStep_(ctx.prevStep)
         , isLast_(ctx.isLast)
-        //@@start issues-to-dst-init
         , issuesToDst_(
               dst == ctx.strandDst && src == ctx.strandDeliver.account &&
               c == ctx.strandDeliver.currency)
-        //@@end issues-to-dst-init
         , j_(ctx.j)
     {
     }
@@ -391,7 +387,6 @@ DirectIOfferCrossingStep::quality(ReadView const&, QualityDirection qDir) const
     return QUALITY_ONE;
 }
 
-//@@start recipient-limit-max-flow
 std::pair<IOUAmount, DebtDirection>
 DirectIPaymentStep::maxFlow(ReadView const& sb, IOUAmount const&) const
 {
@@ -415,9 +410,7 @@ DirectIPaymentStep::maxFlow(ReadView const& sb, IOUAmount const&) const
 
     return {amount, direction};
 }
-//@@end recipient-limit-max-flow
 
-//@@start offer-crossing-precedent
 std::pair<IOUAmount, DebtDirection>
 DirectIOfferCrossingStep::maxFlow(ReadView const& sb, IOUAmount const& desired)
     const
@@ -439,7 +432,6 @@ DirectIOfferCrossingStep::maxFlow(ReadView const& sb, IOUAmount const& desired)
 
     return maxPaymentFlow(sb);
 }
-//@@end offer-crossing-precedent
 
 TER
 DirectIPaymentStep::check(
@@ -481,7 +473,6 @@ DirectIPaymentStep::check(
         }
     }
 
-    //@@start recipient-limit-dry-test
     // The destination's limit does not apply to its issuer's step into it
     // (see maxFlow); a dry test against it would refuse a payment the
     // recipient is allowed to receive.
@@ -499,7 +490,6 @@ DirectIPaymentStep::check(
             }
         }
     }
-    //@@end recipient-limit-dry-test
     return tesSUCCESS;
 }
 
@@ -516,7 +506,6 @@ DirectIOfferCrossingStep::check(
 
 //------------------------------------------------------------------------------
 
-//@@start limit-reader-max-payment-flow
 template <class TDerived>
 std::pair<IOUAmount, DebtDirection>
 DirectStepI<TDerived>::maxPaymentFlow(ReadView const& sb) const
@@ -532,7 +521,6 @@ DirectStepI<TDerived>::maxPaymentFlow(ReadView const& sb) const
         creditLimit2(sb, dst_, src_, currency_) + srcOwed,
         DebtDirection::issues};
 }
-//@@end limit-reader-max-payment-flow
 
 template <class TDerived>
 DebtDirection
--- a/src/xrpld/app/paths/detail/DirectStep.cpp
+++ b/src/xrpld/app/paths/detail/DirectStep.cpp
@@ -44,6 +44,19 @@ protected:
     // Charge transfer fees when the prev step redeems
     Step const* const prevStep_ = nullptr;
     bool const isLast_;
+    //@@start issues-to-dst-member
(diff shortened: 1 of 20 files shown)


src/xrpld/app/paths/detail/DirectStep.cpp
```diff
@@ -99,6 +110,9 @@ class DirectStepI : public StepImp<IOUAmount, IOUAmount, DirectStepI<TDerived>>
         , currency_(c)
         , prevStep_(ctx.prevStep)
         , isLast_(ctx.isLast)
+        , issuesToDst_(
+              dst == ctx.strandDst && src == ctx.strandDeliver.account &&
```
sublimator, line 114: TODO: Separate over-redemption protection from generic-currency receipts. Currently, explicitly naming the issuer bypasses the recipient’s limit, while `Amount.issuer == Destination` still enforces it. Investigate whether both forms can follow the same receipt rule without allowing over-redemption.

Also consider whether the distinction is useful: explicit-issuer payments fix the asset received, while generic-currency payments use trust limits to select acceptable issuers. Compare concrete examples before deciding to unify them.

PR #812, opened 2026-09-16
build: bump boost to 1.91.0 and fix beast/asio timeout bugs

sublimator:
## Summary

Bumps Boost 1.86 → **1.91.0** by backporting XRPLF/rippled #5570 ("Update to Boost 1.88", `1506e65558`) onto `dev`, then the version bump.

**Why:** Boost 1.86 ships a Boost.Beast regression (boostorg/beast#2925 / #2941, fixed in 1.87 by #2926): `basic_stream`'s "timer already expired" branch moves the operation before reading its executor, and under `BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT` (which we set) a moved-from executor is null → `boost::asio::bad_executor` → `std::terminate`. It fires when an HTTP read starts after the 30 s idle timeout (`BaseHTTPPeer::timeoutSeconds`) — i.e. any keep-alive client that idles and resumes. We reproduced it three times on the `export-rng-693` testnet gateway behind caddy (cores + backtrace in the hookz notes); XRPLF hit the same crash on their testnet in Sept 2025 and downgraded their release branches to 1.83 (XRPLF #5759, Bronek's writeup; #5872), while `develop` kept the 1.88 port and ships 1.91 since 3.2.0. We've been on the one version they fled.

1.87 removed the deprecated APIs this tree used (legacy Boost.Coroutine `spawn`, `io_service`, `basic_waitable_timer` getters, `socket_base::max_connections`, `address::from_string`), so it is a port, not a version string — hence carrying #5570.

## Commits (deliberately not squashed, for review)

1. `8bd015d8d8` — raw `cherry-pick -x` of #5570 **with conflict markers left in** (78 files clean, 28 conflicted). Kept so the merge is visible.
2. `b8c80c51db` — the resolution. xahaud-only code around every conflict preserved: UDP door in `ServerImpl`, sticky endpoint selection in `WorkBase` (`strand.wrap` → `bind_executor`), `useSqLiteStore` in `PeerfinderManager`, the #677 `RPCSub`/`HTTPClient` rewrite. `src/xrpld/rpc/detail/RPCSub.cpp` (upstream path) dropped in favour of our `src/xrpld/net/detail/RPCSub.cpp`. Boost 1.88 removes the `io_service` typedef and `address::from_string` and makes Process v2 default: #677's tests move to `io_context`/`executor_work_guard`/`make_address`, `Main.cpp` includes the v1 process headers it already calls.
3. `6e42617c3a` — `ValidatorSite`: `timer_.expires_at()` getter → `expiry()` (as XRPLF); `expires_from_now` → `expires_after` in tests.
4. `fc9cbdf60a` — last `io_service` in `RPCSub::sendThread`.
5. `2c0d679b10` — `boost/1.91.0` (XRPLF's 1.88→1.90→1.91 is version-only for the parts that apply here).
6. `db91bdda03` — 1.91's conan recipe drops the `boost::system` component (header-only): `nudb/2.0.8` → `2.0.9` (its recipe requires `boost::thread`+`boost::headers`, matching XRPLF), and `Boost::system` removed from `cmake/deps/Boost.cmake` as XRPLF did in #6280.
7. `6a32ba2cca` — regression guard `src/test/server/BeastTimeout_test.cpp` (8 cases / 40 checks: raw + composed HTTP reads/writes on and off the TS strand, past deadline). Passes native 1.91 on GCC 11 (hbb image) and Apple Clang 21; **fails on stock 1.86 with `bad_executor`**, so a downgrade or a regression of that path goes red.

8. `3907c33ed7` — **fix for a bug in #5570 itself**: its rewrite of the deadline arming (`expires_from_now(t, ec)` → `try { expires_after(t) } catch {…}`) put the `async_wait(handleDeadline)` *inside the catch*, so the HTTP client's deadline is only registered when arming throws — on the normal path a request to a server that stalls never times out. Caught by #677's `ripple.net.HTTPClient` "Socket cleanup after timeout", which hung the full unit run deterministically on both 1.88 and 1.91 builds (strace: no timer pending on the test's io_context; `handleDeadline` never ran). The wait is back on the success path under `!mShutdown`. **XRPLF `develop` still has the same structure today** (`src/libxrpl/net/HTTPClient.cpp` ~158–170, lambda form) — a live upstream bug affecting `RPCSub` webhook delivery and the CLI's `fromNetwork` client; untested there.

9. `1de02fd043` — `HTTPClient::Query` brace-initialised: it is an aggregate, and `make_shared<Query>(a, b, c)` relies on C++20 parenthesised aggregate init, which CI's clang rejects (GCC 11 accepts it; upstream builds with a newer clang).
10. `4ccb6eb450` — `x-format-changed --since origin/dev`.

11. `af3f2914d4` + `393e08fcf7` + `5a3c9cc093` — coverage suites for the ported plumbing nothing had ever tested: `ripple.insight.StatsDCollector` (loopback UDP sink, every metric kind + hook, teardown), `ripple.basics.ResolverAsio` (numeric host, batch with an unparseable name, `.invalid`, stop with pending work, stop_async), `ripple.server.Spawn` (both overloads, suspension, both exception-propagation branches).
12. `0a71799738` — **more #5570 fallout in `HTTPClient`** (codex): the old `expires_from_now(t, ec)` cleared `mShutdown` on success and `expires_after()` does not, so a failed first host silently skipped the fallback hosts; the deadline now closes the transport instead of waiting on a graceful TLS shutdown a silent peer never answers; `timed_out` is preserved through header completion. Three bounded cases (fallback after refusal, silent TLS peer, stalled TLS handshake).
13. `ca7872fa2d` — `ripple.overlay.OverlayLifecycle` (codex, 7 cases against real `ConnectAttempt`/`PeerImp`, loopback TLS) and a one-line fix for a **pre-existing** gap it exposed: `ConnectAttempt::onWrite` cancelled its timer and read the HTTP upgrade response with no deadline rearmed, so an outbound peer that never answered hung the attempt. Predates #5570.
14. `496224c799` — **`Spawn.h`'s plain-executor branch never compiled** (from #5570): `make_strand(get_associated_executor(ctx))` resolves a plain executor to asio's inline executor, which cannot be stranded; nothing in the tree had instantiated the branch until the new suite did. Now `make_strand(ctx)`.

## Validation so far

- `release-builder.sh` (hbb, GCC 11) on the build box: 752/752, `ldd` glibc-only, `libcheck` ok, at both 1.88 and 1.91.
- Isolated on the tip (GCC 11 hbb build, stock ubuntu), 0 failures: `OverlayLifecycle` 7 cases/51 tests, `HTTPClient` 20/47, `RPCSub` 4/14, `StatsDCollector` 2/3, `ResolverAsio` 5/23, `Spawn` 4/6.
- Full unit suite (`docker-unit-tests.sh`, 48 jobs) on the tip: **293 suites, 3,833 cases, 3,957,237 tests, 0 failures, 156 s**.
- A control build (the tip with the two `HTTPClient`/`ConnectAttempt` product fixes reverted, every test kept) is run to confirm the new cases fail without the fixes; result in the PR conversation.
- Not touched: the `BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT` executor model (XRPLF keeps it on 1.91 too); XRPLF's unrelated dependency bumps (protobuf/sqlite/zlib/jemalloc/libarchive).

## Follow-ups (separate PRs)

- Remove the unused vendored `external/nudb/` recipe (only stale BUILD.md lines reference it).


sublimator:
**Failing-control result** (tip `496224c799` with the two product fixes reverted — `HTTPClient.cpp` before `0a71799738`, `ConnectAttempt.cpp` before `ca7872fa2d` — every test suite kept; GCC 11 hbb build, stock ubuntu):

- `ripple.net.HTTPClient`: **8 failures, all inside the three new cases** — "Try the next host after connection refused" (4 checks: the un-reset `mShutdown` skipping fallback hosts), "Deadline while TLS peer never responds" (2), "Deadline during TLS handshake" (2). The other 17 cases pass on both binaries.
- `ripple.overlay.OverlayLifecycle`: **3 failures, all in "stalled handshake and HTTP response expire"** — the `ConnectAttempt::onWrite` deadline rearm. The other 6 cases pass on both binaries.

On the tip itself: both suites 0 failures; full suite 293 suites / 3,833 cases / 3,957,237 tests / 0 failures (156 s, 48 jobs).

sublimator:
TODO / evidence of shutdown hangs on this branch: on a 4-node testnet running dev + this PR (+ feature-export-rng, merge `df51b8ef5f`), every node sent Ctrl-C hung in shutdown with

```
JobQueue:ERR JobQueue waiting for ClosureCounter::join().
```

still holding its ports, until `kill -9`. Nodes had ~30 min uptime and open ws subscriptions (`export_signatures`, `ledger`) at the time. Not yet isolated to the Boost bump vs the feature branch; a vanilla dev+#812 net would tell.

-- 

Later run[s] could not reproduce, and sadly no origin/dev baseline for this behaviour 
Inconclusive

sublimator:
```
  
  
  boost/1.91.0: Package '1d978fc97431de2d1317bf1e72ebedb40f4b95b6' built
  boost/1.91.0: Build folder /Users/mac-ci-001/.conan2/p/b/boost99c36df8a3b2c/b/build-debug
  >>> Package step for boost/1.91.0:1d978fc97431de2d1317bf1e72ebedb40f4b95b6
  Packaging in folder /Users/mac-ci-001/.conan2/p/b/boost99c36df8a3b2c/p
  Calling package()
  package(): Packaged 1 '.txt' file: LICENSE_1_0.txt
  package(): Packaged 15855 '.hpp' files
  package(): Packaged 151 '.h' files
  package(): Packaged 307 '.ipp' files
  package(): Packaged 2 '.inc' files: strict_cpp_re.inc, cpp_re.inc
  package(): Packaged 46 '.a' files
  Created package revision 7230c898b31ad0b53b886b2bc9766e82
  Package '1d978fc97431de2d1317bf1e72ebedb40f4b95b6' created
  Full package reference: boost/1.91.0#5a53ec91421b101535a37c248a5654fc:1d978fc97431de2d1317bf1e72ebedb40f4b95b6#7230c898b31ad0b53b886b2bc9766e82
  Package folder /Users/mac-ci-001/.conan2/p/b/boost99c36df8a3b2c/p
  boost/1.91.0: WARN: Boost component 'cobalt_io_ssl' is missing libraries. Try building boost with '-o boost:without_cobalt_io_ssl'. (Option is not guaranteed to exist)
  ERROR: boost/1.91.0: Error in package_info() method, line 2044
  	raise ConanException(f"These libraries were expected to be built, but were not built: {non_built}")
  	ConanException: These libraries were expected to be built, but were not built: {'boost_cobalt_io_ssl'}
  Error: Process completed with exit code 1.
0s

```

